PR #6, opened 2020-03-05
feat(cv): add cstor volume manager/controller using new APIs

prateekpandey14:
changes:
- add cstor volume manager/controller using new APIs …
-  remove maya as vendor dependency


pkg/controllers/volume-mgmt/controller.go
```diff
@@ -0,0 +1,139 @@
+/*
+Copyright 2018 The OpenEBS Authors.
```
sonasingh46: n_p: 2020

pkg/controllers/volume-mgmt/controller_test.go
```diff
@@ -0,0 +1,48 @@
+// Copyright © 2017-2019 The OpenEBS Authors
```
sonasingh46: n_p: 2020 
Similarly, at other places.

prateekpandey14: done

pkg/controllers/volume-mgmt/controller.go
prateekpandey14: done

pkg/controllers/volume-mgmt/controller_test.go
prateekpandey14: Update the year in other files as well

PR #14, opened 2020-03-12
refact(webhook): add validation for new fields in cspc

shubham14bajpai:

This PR
-  adds validation for new schema changes in CSPC. 
-  reverts the API version of `ValidatingWebhookConfig` to v1beta1 as the `sideEffectClass` `NoneOnDryRun` does not seem to work in this case and v1 does not support `Unknown` class. Will look into it until then using v1beta1.

TODO:
 - validation for expansion
 - validation for capacity

will take these in separate PRs.

### DIFF
--- a/pkg/webhook/cspc.go
+++ b/pkg/webhook/cspc.go
@@ -247,16 +247,9 @@ func (poolValidator *PoolValidator) poolSpecValidation() (bool, string) {
 	if len(poolValidator.poolSpec.DataRaidGroups) == 0 {
 		return false, "at least one raid group should be present on dataRaidGroups"
 	}
-	if len(poolValidator.poolSpec.WriteCacheRaidGroups) == 0 {
-		return false, "at least one raid group should be present on writeCacheRaidGroups"
-	}
 	if poolValidator.poolSpec.PoolConfig.DataRaidGroupType == string(cstor.PoolStriped) &&
 		len(poolValidator.poolSpec.DataRaidGroups) != 1 {
 		return false, "stripe dataRaidGroups should have exactly one raidGroup"
-	}
-	if poolValidator.poolSpec.PoolConfig.WriteCacheGroupType == string(cstor.PoolStriped) &&
-		len(poolValidator.poolSpec.WriteCacheRaidGroups) != 1 {
-		return false, "stripe writeCacheRaidGroups should have exactly one raidGroup"
 	}
 	for _, raidGroup := range poolValidator.poolSpec.DataRaidGroups {
 		raidGroup := raidGroup // pin it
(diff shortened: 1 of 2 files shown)


pkg/webhook/cspc.go
```diff
@@ -262,11 +318,17 @@ var (
...
+		cstor.PoolStriped:  isStripedBDCountValid,
+		cstor.PoolMirrored: isMirroredBDCountValid,
+		cstor.PoolRaidz:    isRaidzBDCountValid,
+		cstor.PoolRaidz2:   isRaidz2BDCountValid,
+	}
+	// SupportedCompression is a map holding the supported compressions
+	SupportedCompression = map[string]bool{
+		"":    true,
+		"off": true,
+		"lz":  true,
```
mittachaitu: Not sure what and all are supporting
```c
 static zprop_index_t compress_table[] = {
                { "on",         ZIO_COMPRESS_ON },
                { "off",        ZIO_COMPRESS_OFF },
                { "lzjb",       ZIO_COMPRESS_LZJB },
                { "gzip",       ZIO_COMPRESS_GZIP_6 },  /* gzip default */
                { "gzip-1",     ZIO_COMPRESS_GZIP_1 },
                { "gzip-2",     ZIO_COMPRESS_GZIP_2 },
                { "gzip-3",     ZIO_COMPRESS_GZIP_3 },
                { "gzip-4",     ZIO_COMPRESS_GZIP_4 },
                { "gzip-5",     ZIO_COMPRESS_GZIP_5 },
                { "gzip-6",     ZIO_COMPRESS_GZIP_6 },
                { "gzip-7",     ZIO_COMPRESS_GZIP_7 },
                { "gzip-8",     ZIO_COMPRESS_GZIP_8 },
                { "gzip-9",     ZIO_COMPRESS_GZIP_9 },
                { "zle",        ZIO_COMPRESS_ZLE },
                { "lz4",        ZIO_COMPRESS_LZ4 },
                { NULL }
        };
```
cc: @sonasingh46

mittachaitu: @shubham14bajpai We need to support above all compressions.

mittachaitu: As of know, the existing map is good. We will revisit if we need to support more...

shubham14bajpai: Added a TODO comment for this. Once we are sure we can update the map.

pkg/webhook/cspc.go
```diff
@@ -285,17 +362,11 @@ func (poolValidator *PoolValidator) raidGroupValidation(
 	raidGroup *cstor.RaidGroup, rgType string) (bool, string) {
 
 	if len(raidGroup.BlockDevices) == 0 {
-		return false, fmt.Sprintf("number of block devices honouring raid type should be specified")
```
mittachaitu: Below check is good enough we don't require this check(We can remove in next PR).

PR #15, opened 2020-03-12
feat(roThresholdLimit): add ROThresholdLimit support in pool manager

mittachaitu:
This PR adds support for ROThresholdLimit in the pool manager.

Note:
1. If `roThresholdLimit` is set to `0` or `100`:
    -  Writes on the pool will be allowed until it runs out of space.

Sample CSPI spec and status when pool in RW state
```sh
    poolConfig:
      ...
      ...
      roThresholdLimit: 12
  status:
    Conditions: null
    capacity:
      free: 9470M
      total: 9940M
      used: 476M
    phase: ""
    readOnly: false
```
Sample CSPI spec and status when pool is in RO mode
```sh
    poolConfig:
      ...
      ...
      roThresholdLimit: 2
  status:
    Conditions: null
    capacity:
      free: 9470M
      total: 9940M
      used: 476M
    phase: ""
    readOnly: true
```
Events Generation:

```sh
I0312 18:00:30.985242   19525 event.go:281] Event(v1.ObjectReference{Kind:"CStorPoolInstance", Namespace:"openebs", Name:"cspi-stripe", UID:"be8e517b-508d-4044-93bd-67e668593d4d", APIVersion:"cstor.openebs.io/v1", ResourceVersion:"1923", FieldPath:""}): type: 'Warning' reason: 'PoolReadOnlyThreshold' Pool storage limit reached to thresholdLimit. Pool expansion is required to make it's volume replicas RW
...
...
I0312 18:31:55.654584   26872 event.go:281] Event(v1.ObjectReference{Kind:"CStorPoolInstance", Namespace:"openebs", Name:"cspi-stripe", UID:"be8e517b-508d-4044-93bd-67e668593d4d", APIVersion:"cstor.openebs.io/v1", ResourceVersion:"3172", FieldPath:""}): type: 'Normal' reason: 'PoolReadOnlyThreshold' Pool roThresholdLimit or pool got expanded due to that pool readOnly mode is unset
```
TODO:
Fix travis after PR in api got merged.

### DIFF
--- a/pkg/controllers/cspi-controller/handler.go
+++ b/pkg/controllers/cspi-controller/handler.go
@@ -195,40 +195,26 @@ func (c *CStorPoolInstanceController) updateStatus(cspi *cstor.CStorPoolInstance
 	var status cstor.CStorPoolInstanceStatus
 	var err error
 	pool := zpool.PoolName()
+	propertyList := []string{"health", "io.openebs:readonly"}
 
-	state, er := zpool.GetPropertyValue(pool, "health")
-	if er != nil {
-		err = zpool.ErrorWrapf(err, "Failed to fetch health")
+	valueList, er := zpool.GetListOfPropertyValues(pool, propertyList)
+	if er != nil || len(valueList) != len(propertyList) {
+		return errors.Wrapf(err, "Failed to fetch %v output: %v", propertyList, valueList)
 	} else {
-		status.Phase = cstor.CStorPoolInstancePhase(state)
+		status.Phase = cstor.CStorPoolInstancePhase(valueList[0])
+		if valueList[1] == "on" {
+			status.ReadOnly = true
+		}
 	}
 
-	freeSize, er := zpool.GetPoolCapacity(pool, "free")
-	if er != nil {
-		err = zpool.ErrorWrapf(err, "Failed to fetch free size")
-	} else {
-		status.Capacity.Free = freeSize
-	}
-
-	usedSize, er := zpool.GetPoolCapacity(pool, "allocated")
-	if er != nil {
-		err = zpool.ErrorWrapf(err, "Failed to fetch used size")
-	} else {
-		status.Capacity.Used = usedSize
-	}
-
-	totalSize, er := zpool.GetPoolCapacity(pool, "size")
-	if er != nil {
-		err = zpool.ErrorWrapf(err, "Failed to fetch total size")
-	} else {
-		status.Capacity.Total = totalSize
-	}
-
+	status.Capacity, err = zpool.GetCSPICapacity(pool)
 	if err != nil {
 		return errors.Errorf("Failed to sync due to %s", err.Error())
 	}
+	c.updateROMode(&status, *cspi)
 
 	if IsStatusChange(cspi.Status, status) {
+		klog.Infof("Status %v", status)
 		cspi.Status = status
 		_, err = zpool.OpenEBSClient.
 			CstorV1().
@@ -239,6 +225,51 @@ func (c *CStorPoolInstanceController) updateStatus(cspi *cstor.CStorPoolInstance
 		}
 	}
 	return nil
+}
+
+// updateROMode sets/unsets the pool readonly mode property. It does the following changes
+// 1. If pool used space reached to roThresholdLimit then pool will be set to readonly mode
+// 2. If pool was in readonly mode if roThresholdLimit/pool expansion was happened then it
+//    unsets the ReadOnly Mode.
+// NOTE: This function must be invoked after having the updated cspiStatus information
+func (c *CStorPoolInstanceController) updateROMode(
+	cspiStatus *cstor.CStorPoolInstanceStatus, cspi cstor.CStorPoolInstance) {
+	roThresholdLimit := cspi.Spec.PoolConfig.ROThresholdLimit
+	totalInBytes := cspiStatus.Capacity.Total.Value()
+	usedInBytes := cspiStatus.Capacity.Used.Value()
+	pool := zpool.PoolName()
+
+	usedPercentage := (usedInBytes * 100) / totalInBytes
+	if (int(usedPercentage) >= roThresholdLimit) &&
+		(roThresholdLimit != 0 && roThresholdLimit != 100) {
+		if !cspiStatus.ReadOnly {
+			if err := zpool.SetPoolRDMode(pool, true); err != nil {
+				// Here, we are just logging in next reconciliation it will be retried
+				klog.Errorf("failed to set pool ReadOnly Mode to %t error: %s", true, err.Error())
+			} else {
+				cspiStatus.ReadOnly = true
+				c.recorder.Event(&cspi,
+					corev1.EventTypeWarning,
+					"PoolReadOnlyThreshold",
+					"Pool storage limit reached to thresholdLimit. Pool expansion is required to make it's volume replicas RW",
+				)
+			}
+		}
+	} else {
+		if cspiStatus.ReadOnly {
+			if err := zpool.SetPoolRDMode(pool, false); err != nil {
+				klog.Errorf("Failed to unset pool readOnly mode : %v", err)
+			} else {
+				cspiStatus.ReadOnly = false
+				c.recorder.Event(&cspi,
+					corev1.EventTypeNormal,
+					"PoolReadOnlyThreshold",
+					"Pool roThresholdLimit or pool got expanded due to that pool readOnly mode is unset",
+				)
+			}
+
+		}
+	}
 }
 
 // getCSPIObjFromKey returns object corresponding to the resource key
--- a/pkg/controllers/cspi-controller/handler.go
+++ b/pkg/controllers/cspi-controller/handler.go
@@ -239,10 +239,10 @@ func (c *CStorPoolInstanceController) updateStatus(cspi *cstor.CStorPoolInstance
 //       cspiStatus information from zfs/zpool
 func (c *CStorPoolInstanceController) updateROMode(
 	cspiStatus *cstor.CStorPoolInstanceStatus, cspi cstor.CStorPoolInstance) {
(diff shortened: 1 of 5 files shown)


pkg/controllers/cspi-controller/handler.go
```diff
@@ -195,40 +195,26 @@ func (c *CStorPoolInstanceController) updateStatus(cspi *cstor.CStorPoolInstance
...
 	var err error
 	pool := zpool.PoolName()
+	propertyList := []string{"health", "io.openebs:readonly"}
 
-	state, er := zpool.GetPropertyValue(pool, "health")
-	if er != nil {
-		err = zpool.ErrorWrapf(err, "Failed to fetch health")
+	valueList, er := zpool.GetListOfPropertyValues(pool, propertyList)
+	if er != nil || len(valueList) != len(propertyList) {
+		return errors.Wrapf(err, "Failed to fetch %v output: %v", propertyList, valueList)
```
sonasingh46: Is the `valueList` and  `propertyList` in order ? 
I mean if let say , propertyList is -- ["Prop1", "Prop2"]
Will the valueList be in respective order -- ["valueForProp1", "valueForProp2"] ?

pkg/controllers/cspi-controller/handler.go
```diff
@@ -195,40 +195,26 @@ func (c *CStorPoolInstanceController) updateStatus(cspi *cstor.CStorPoolInstance
...
-
-	totalSize, er := zpool.GetPoolCapacity(pool, "size")
-	if er != nil {
-		err = zpool.ErrorWrapf(err, "Failed to fetch total size")
-	} else {
-		status.Capacity.Total = totalSize
+		status.Phase = cstor.CStorPoolInstancePhase(valueList[0])
+		if valueList[1] == "on" {
+			status.ReadOnly = true
+		}
```
sonasingh46: What are the magical numbers `0` and `1` ? 
If we need to use them -- a little comment can help understand.

pkg/controllers/cspi-controller/handler.go
```diff
@@ -241,6 +227,51 @@ func (c *CStorPoolInstanceController) updateStatus(cspi *cstor.CStorPoolInstance
 	return nil
 }
 
+// updateROMode sets/unsets the pool readonly mode property. It does the following changes
+// 1. If pool used space reached to roThresholdLimit then pool will be set to readonly mode
+// 2. If pool was in readonly mode if roThresholdLimit/pool expansion was happened then it
+//    unsets the ReadOnly Mode.
+// NOTE: This function must be invoked after having the updated cspiStatus information
```
sonasingh46: :100: for comments...  
Trying to understand the `NOTE` section of comment...

pkg/controllers/cspi-controller/handler.go
```diff
@@ -241,6 +227,51 @@ func (c *CStorPoolInstanceController) updateStatus(cspi *cstor.CStorPoolInstance
...
+func (c *CStorPoolInstanceController) updateROMode(
+	cspiStatus *cstor.CStorPoolInstanceStatus, cspi cstor.CStorPoolInstance) {
+	roThresholdLimit := cspi.Spec.PoolConfig.ROThresholdLimit
+	totalInBytes := cspiStatus.Capacity.Total.Value()
+	usedInBytes := cspiStatus.Capacity.Used.Value()
+	pool := zpool.PoolName()
+
+	usedPercentage := (usedInBytes * 100) / totalInBytes
+	if (int(usedPercentage) >= roThresholdLimit) &&
+		(roThresholdLimit != 0 && roThresholdLimit != 100) {
```
sonasingh46: I will rather say -- 
1. `roThresholdLimit==0` should mark pool as read only. 
Me as a user -- I want to set the pool to read only (why ? ) . How can I do that ? 
Well edit your cspc and put `roThreshold` to 0.

Then, how we differentiate b/w 0 `roThreshold` and not specified  `roThreshold`
Well then -- lets make  ` roThreshold` a pointer field.

2. And `roThresholdLimit==100` by defaults fits into maths for putting pool into RO ? 
is there a situation when the pool is exhausted 100% and still we do not want to put it in RO as we do not support unmap and want to rewrite ?

pkg/controllers/cspi-controller/handler.go
```diff
@@ -241,6 +227,51 @@ func (c *CStorPoolInstanceController) updateStatus(cspi *cstor.CStorPoolInstance
...
+		if !cspiStatus.ReadOnly {
+			if err := zpool.SetPoolRDMode(pool, true); err != nil {
+				// Here, we are just logging in next reconciliation it will be retried
+				klog.Errorf("failed to set pool ReadOnly Mode to %t error: %s", true, err.Error())
+			} else {
+				cspiStatus.ReadOnly = true
+				c.recorder.Event(&cspi,
+					corev1.EventTypeWarning,
+					"PoolReadOnlyThreshold",
+					"Pool storage limit reached to thresholdLimit. Pool expansion is required to make it's volume replicas RW",
```
sonasingh46: n_p: 
`"Pool storage limit reached to read only thresholdLimit. Pool expansion is required to make its volume replicas RW"`
Notice the -- `it's --> its`

pkg/controllers/cspi-controller/handler.go
```diff
@@ -241,6 +227,51 @@ func (c *CStorPoolInstanceController) updateStatus(cspi *cstor.CStorPoolInstance
...
+	cspiStatus *cstor.CStorPoolInstanceStatus, cspi cstor.CStorPoolInstance) {
+	roThresholdLimit := cspi.Spec.PoolConfig.ROThresholdLimit
+	totalInBytes := cspiStatus.Capacity.Total.Value()
+	usedInBytes := cspiStatus.Capacity.Used.Value()
+	pool := zpool.PoolName()
+
+	usedPercentage := (usedInBytes * 100) / totalInBytes
+	if (int(usedPercentage) >= roThresholdLimit) &&
+		(roThresholdLimit != 0 && roThresholdLimit != 100) {
+		if !cspiStatus.ReadOnly {
```
sonasingh46: What is status is ReadOnly but pool is not readOnly ? I mean this could be a bad state but if we get into this state -- we won't be able to recover. 
Ans these things can happen due to API server unreachability.
Pool if needs to go into RO should go into RO even if the status is stale...
The idea is finally things should converge towards a good guy.

pkg/controllers/cspi-controller/handler.go
```diff
@@ -241,6 +227,51 @@ func (c *CStorPoolInstanceController) updateStatus(cspi *cstor.CStorPoolInstance
...
+					corev1.EventTypeNormal,
+					"PoolReadOnlyThreshold",
+					"Pool roThresholdLimit or pool got expanded due to that pool readOnly mode is unset",
+				)
+			}
+
+		}
+	}
+}
+
```
sonasingh46: Similar comment as above... 
Pool should go into RO if it needs to go and pool should go into RW if it needs to go irrespective of status...
Finally, the status can be updated with RO thing.
Happy to learn more and discuss.

pkg/cspc/algorithm/build_csp.go
```diff
@@ -58,6 +58,10 @@ func (ac *Config) GetCSPISpec() (*cstor.CStorPoolInstance, error) {
 		poolSpec.PoolConfig.PriorityClassName = ac.CSPC.Spec.DefaultPriorityClassName
 	}
 
+	if poolSpec.PoolConfig.ROThresholdLimit == 0 {
+		poolSpec.PoolConfig.ROThresholdLimit = 85
+	}
+
```
sonasingh46: If we convert `ROThresholdLimit` to a pointer -- it would then become non-ambiguous to say intentional zero value of a field.

pkg/controllers/cspi-controller/handler.go
mittachaitu: Yes!! It will be queried in the same order

pkg/controllers/cspi-controller/handler.go
mittachaitu: NOTE: Is trying to convey that this function should be called only after getting the updated info from the pool(by using zpool/zfs command). If we call this function with stale/outdated capacity values then the behaviour will not be as expected.

pkg/controllers/cspi-controller/handler.go
mittachaitu: and output also will be in the same order

pkg/controllers/cspi-controller/handler.go
mittachaitu: > roThresholdLimit==0 should mark pool as read only.
Me as a user -- I want to set the pool to read only (why ? ) . How can I do that ?
Well edit your cspc and put roThreshold to 0.

I also thought of treating it as a pointer but I am not aware of any such use case to honor zero value of roThresholdLimit.

mittachaitu: > And roThresholdLimit==100 by defaults fits into maths for putting pool into RO ?
is there a situation when the pool is exhausted 100% and still we do not want to put it in RO as we do not support unmap and want to rewrite ?

If the user wants to use the entier pool we shouldn't stop the user but anyhow, in reality, 100% pool can't be consumed even though if it consumed we can't perform any operations on the pool until expansion happens. So If user-configured 100% we shouldn't touch the pool will be automatically transferred to some other state(something like non-operational mode).

pkg/controllers/cspi-controller/handler.go
mittachaitu: We will not be dealing with stale/outdated info since we are getting info from the pool by doing zfs/zpool commands before calling this function
https://github.com/openebs/cstor-operators/blob/e1d33c5449ff0133b09441463c111428a67d02ae/pkg/controllers/cspi-controller/handler.go#L206

pkg/cspc/algorithm/build_csp.go
mittachaitu: Done

PR #17, opened 2020-03-17
chore(cvc): refactor volume and pool provisioning including updated image names

prateekpandey14:
- fix volume provisioning due to stale error checks
- Rename `pool-manager` image to `cstor-pool-manager`
- rename `volume-manager` image to `cstor-volume-manager`
- Use `IfNotPresent` instead of `Always` pull container images in pool deployment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,7 +97,7 @@ volume-manager-image:
 	@echo "----------------------------"
 	@PNAME=${VOLUME_MANAGER} CTLNAME=${VOLUME_MANAGER} sh -c "'$(PWD)/build/build.sh'"
 	@cp bin/${VOLUME_MANAGER}/${VOLUME_MANAGER} build/volume-manager/
-	@cd build/${VOLUME_MANAGER} && sudo docker build -t ${HUB_USER}/${VOLUME_MANAGER}:${IMAGE_TAG} --build-arg BUILD_DATE=${BUILD_DATE} .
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -97,7 +97,7 @@ volume-manager-image:
 	@echo "----------------------------"
 	@PNAME=${VOLUME_MANAGER} CTLNAME=${VOLUME_MANAGER} sh -c "'$(PWD)/build/build.sh'"
 	@cp bin/${VOLUME_MANAGER}/${VOLUME_MANAGER} build/volume-manager/
-	@cd build/${VOLUME_MANAGER} && sudo docker build -t ${HUB_USER}/${VOLUME_MANAGER}:${IMAGE_TAG} --build-arg BUILD_DATE=${BUILD_DATE} .
+	@cd build/${VOLUME_MANAGER} && sudo docker build -t ${HUB_USER}/cstor-${VOLUME_MANAGER}:${IMAGE_TAG} --build-arg BUILD_DATE=${BUILD_DATE} .
```
mittachaitu: Better to update the VOLUME_MANAGER with cstor-volume-manager

shubham14bajpai: The VOLUME_MANAGER variable is the directory name and binary name, IMO it should not be changed to cstor-volume-manager as the parent directory has cstor in it. We can a have a separate variable VOLUME_MANAGER_IMAGE as cstor-volume-manager.

PR #19, opened 2020-03-18
feat(cspc): add reconciliation for cspc tunables

sonasingh46:
This PR adds functionality to reconcile for cspc tunables.

### DIFF
--- a/pkg/controllers/cspc-controller/sync.go
+++ b/pkg/controllers/cspc-controller/sync.go
@@ -267,13 +267,13 @@ func (c *Controller) GetCSPIWithoutDeployment(cspc *cstor.CStorPoolCluster) ([]c
 }
 
 // syncCSPI propagates all the required changes from cspc to respective cspi.
-func (pc *PoolConfig) syncCSPI(cspc *cstor.CStorPoolCluster) {
+func (pc *PoolConfig) syncCSPI(cspc *cstor.CStorPoolCluster) error {
 	cspiList, err := pc.Controller.GetCSPIListForCSPC(cspc)
 	if err != nil {
-		klog.Errorf("failed to sync cspi(s) from its parent cspc %s", cspc.Name)
+		return errors.Wrapf(err, "failed to sync cspi(s) from its parent cspc %s", cspc.Name)
 	}
 	if len(cspiList.Items) == 0 {
-		klog.Errorf("No cspi(s) found while trying to sync cspi(s) from its parent cspc %s", cspc.Name)
+		return errors.Wrapf(err, "No cspi(s) found while trying to sync cspi(s) from its parent cspc %s", cspc.Name)
 	}
 
 	for _, cspi := range cspiList.Items {
@@ -294,7 +294,7 @@ func (pc *PoolConfig) syncCSPIWithCSPC(cspc *cstor.CStorPoolCluster, cspi *cstor
 		defaultPoolConfig(cspiCopy, cspc)
 	}
 
-	if reflect.DeepEqual(cspiCopy, cspi) {
+	if !reflect.DeepEqual(cspiCopy, cspi) {
 		gotCSPI, err := pc.Controller.GetStoredCStorVersionClient().CStorPoolInstances(cspiCopy.Namespace).Update(cspiCopy)
 		if err != nil {
 			return errors.Errorf("Failed to sync cspi %s from parent cspc %s", cspiCopy.Name, cspc.Name)
--- a/pkg/controllers/cspc-controller/sync.go
+++ b/pkg/controllers/cspc-controller/sync.go
@@ -17,15 +17,20 @@ limitations under the License.
 package cspccontroller
 
 import (
+	"encoding/json"
 	"fmt"
 	cstor "github.com/openebs/api/pkg/apis/cstor/v1"
 	"github.com/openebs/api/pkg/apis/types"
 	"github.com/openebs/cstor-operators/pkg/cspc/algorithm"
 	"github.com/openebs/maya/pkg/version"
 	"github.com/pkg/errors"
+	"k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	k8serror "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8stypes "k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/mergepatch"
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	"k8s.io/klog"
 	"reflect"
 )
@@ -108,7 +113,12 @@ func (c *Controller) sync(cspc *cstor.CStorPoolCluster, cspiList *cstor.CStorPoo
 	}
 
 	// sync changes to cspi from cspc e.g. tunables like toleration, resource requirements etc
-	pc.syncCSPI(cspc)
+	err = pc.syncCSPI(cspc)
+
+	// Not returning error so that `handleOperations` can also be executed.
+	if err != nil {
+		klog.Errorf("failed to sync cspi(s) of cspc %s", cspc.Name)
+	}
 
 	pc.handleOperations()
 	return nil
@@ -266,7 +276,12 @@ func (c *Controller) GetCSPIWithoutDeployment(cspc *cstor.CStorPoolCluster) ([]c
 	return cspiList, nil
 }
 
-// syncCSPI propagates all the required changes from cspc to respective cspi.
+/*
+syncCSPI propagates all the required changes from cspc to respective cspi.
+ToDo: Currently -- in every resync interval the sync is tried and this needs to be improved by queuing cspc
+only at times when it is required.
+ */
+
 func (pc *PoolConfig) syncCSPI(cspc *cstor.CStorPoolCluster) error {
 	cspiList, err := pc.Controller.GetCSPIListForCSPC(cspc)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


pkg/controllers/cspc-controller/sync.go
```diff
@@ -261,3 +265,60 @@ func (c *Controller) GetCSPIWithoutDeployment(cspc *cstor.CStorPoolCluster) ([]c
...
+		if err != nil {
+			klog.Errorf("failed to sync cspi %s from its parent cspc %s", cspi.Name, cspc.Name)
+		}
+	}
+}
+
+func (pc *PoolConfig) syncCSPIWithCSPC(cspc *cstor.CStorPoolCluster, cspi *cstor.CStorPoolInstance) error {
+	cspiCopy := cspi.DeepCopy()
+	klog.V(2).Infof("Syncing cspi %s from parent cspc %s", cspiCopy.Name, cspc.Name)
+	for _, poolSpec := range cspc.Spec.Pools {
```
mittachaitu: How do we know that pool spec belongs to current CSPI?

pkg/controllers/cspc-controller/sync.go
```diff
@@ -261,3 +265,60 @@ func (c *Controller) GetCSPIWithoutDeployment(cspc *cstor.CStorPoolCluster) ([]c
...
+		poolSpec := poolSpec
+		cspiCopy.Spec.PoolConfig = poolSpec.PoolConfig
+		defaultPoolConfig(cspiCopy, cspc)
+	}
+
+	if reflect.DeepEqual(cspiCopy, cspi) {
```
mittachaitu: We should do only
```go
if !reflect.DeepEqual(cspiCopy, cspi) {
// ...
//...
}`

File: pkg/controllers/cspc-controller/sync.go [hunk 3]
```diff
@@ -261,3 +265,60 @@ func (c *Controller) GetCSPIWithoutDeployment(cspc *cstor.CStorPoolCluster) ([]c
...
+		gotCSPI, err := pc.Controller.GetStoredCStorVersionClient().CStorPoolInstances(cspiCopy.Namespace).Update(cspiCopy)
+		if err != nil {
+			return errors.Errorf("Failed to sync cspi %s from parent cspc %s", cspiCopy.Name, cspc.Name)
+		}
+		_, err = pc.Controller.kubeclientset.AppsV1().Deployments(gotCSPI.Namespace).Update(pc.AlgorithmConfig.GetPoolDeploySpec(gotCSPI))
```
mittachaitu: Is it good to do patch instead of updating deployment(It will lose user-configured labels or annotations -- before that is it Valid use case)?

pkg/controllers/cspc-controller/sync.go
```diff
@@ -261,3 +265,60 @@ func (c *Controller) GetCSPIWithoutDeployment(cspc *cstor.CStorPoolCluster) ([]c
 	}
 	return cspiList, nil
 }
+
+// syncCSPI propagates all the required changes from cspc to respective cspi.
+func (pc *PoolConfig) syncCSPI(cspc *cstor.CStorPoolCluster) {
+	cspiList, err := pc.Controller.GetCSPIListForCSPC(cspc)
+	if err != nil {
+		klog.Errorf("failed to sync cspi(s) from its parent cspc %s", cspc.Name)
```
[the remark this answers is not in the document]
sonasingh46: thanks.

pkg/controllers/cspc-controller/sync.go
sonasingh46: yes.

pkg/controllers/cspc-controller/sync.go
sonasingh46: But do we want to make the pool-manager deployment to be user tunable separately? 
In fact, I thought about this and choose to go with `update` instead of `patch` intentionally.
Thinking on lines of gitops. 
Let us see what @kmova and @vishnuitta have to say about this. 
If we need to make sure that user configured labels/annotations are not lost -- I can do it a strategic merge patch.

IMO -- the pool-manger config should be only dependent on cspi-config and must not be tuned externally but your point of not losing user set annotations and labels also look reasonable. Can discuss and finalize this.

pkg/controllers/cspc-controller/sync.go
```diff
@@ -106,6 +107,14 @@ func (c *Controller) sync(cspc *cstor.CStorPoolCluster, cspiList *cstor.CStorPoo
 		pc.createDeployForCSPList(cspc, cspisWithoutDeployment)
 	}
 
+	// sync changes to cspi from cspc e.g. tunables like toleration, resource requirements etc
+	err=pc.syncCSPI(cspc)
+
+	// Not returning error so that `handleOperations` can also be executed.
+	if err!=nil{
```
prateekpandey14: fmt changes required , run gofmt please

pkg/controllers/cspc-controller/sync.go
```diff
@@ -106,6 +107,14 @@ func (c *Controller) sync(cspc *cstor.CStorPoolCluster, cspiList *cstor.CStorPoo
 		pc.createDeployForCSPList(cspc, cspisWithoutDeployment)
 	}
 
+	// sync changes to cspi from cspc e.g. tunables like toleration, resource requirements etc
+	err=pc.syncCSPI(cspc)
```
prateekpandey14: fmt changes required , run gofmt please

pkg/controllers/cspc-controller/sync.go
```diff
@@ -261,3 +270,63 @@ func (c *Controller) GetCSPIWithoutDeployment(cspc *cstor.CStorPoolCluster) ([]c
...
+		}
+	}
+	return nil
+}
+
+func (pc *PoolConfig) syncCSPIWithCSPC(cspc *cstor.CStorPoolCluster, cspi *cstor.CStorPoolInstance) error {
+	cspiCopy := cspi.DeepCopy()
+	klog.V(2).Infof("Syncing cspi %s from parent cspc %s", cspiCopy.Name, cspc.Name)
+	for _, poolSpec := range cspc.Spec.Pools {
+		if reflect.DeepEqual(poolSpec.NodeSelector,cspi.Spec.NodeSelector){
```
mittachaitu: By seeing this check looks like we shouldn't allow updating or modifying node selector(Just gathering Info).

pkg/controllers/cspc-controller/sync.go
```diff
@@ -261,3 +270,63 @@ func (c *Controller) GetCSPIWithoutDeployment(cspc *cstor.CStorPoolCluster) ([]c
...
+			poolSpec := poolSpec
+			cspiCopy.Spec.PoolConfig = poolSpec.PoolConfig
+			defaultPoolConfig(cspiCopy, cspc)
```
mittachaitu: add break here

pkg/cspc/algorithm/select_node.go
```diff
@@ -43,7 +43,7 @@ func (ac *Config) SelectNode() (*cstor.PoolSpec, string, error) {
 		pool := pool
 		nodeName, err := ac.GetNodeFromLabelSelector(pool.NodeSelector)
 		if err != nil || nodeName == "" {
-			klog.Errorf("could not use node for selectors {%v}", pool.NodeSelector)
+			klog.Errorf("could not use node for selectors {%v}: {%s}", pool.NodeSelector, err)
```
mittachaitu: Let us use format specifier for error as %v (or) change argument from err.Error()

pkg/controllers/cspc-controller/sync.go
sonasingh46: done

pkg/controllers/cspc-controller/sync.go
```diff
@@ -17,16 +17,22 @@ limitations under the License.
...
 
 import (
+	"encoding/json"
 	"fmt"
 	cstor "github.com/openebs/api/pkg/apis/cstor/v1"
 	"github.com/openebs/api/pkg/apis/types"
 	"github.com/openebs/cstor-operators/pkg/cspc/algorithm"
 	"github.com/openebs/maya/pkg/version"
 	"github.com/pkg/errors"
+	"k8s.io/api/apps/v1"
```
mittachaitu: nit: Good to alias it as k8sapps

pkg/controllers/cspc-controller/sync.go
```diff
@@ -261,3 +275,137 @@ func (c *Controller) GetCSPIWithoutDeployment(cspc *cstor.CStorPoolCluster) ([]c
...
+	klog.Infof("Syncing cspi %s from parent cspc %s", cspiCopy.Name, cspc.Name)
+	for _, poolSpec := range cspc.Spec.Pools {
+		if reflect.DeepEqual(poolSpec.NodeSelector, cspi.Spec.NodeSelector) {
+			poolSpec := poolSpec
+			cspiCopy.Spec.PoolConfig = poolSpec.PoolConfig
+			defaultPoolConfig(cspiCopy, cspc)
+			break
+		}
+	}
+	gotCSPI, err := pc.Controller.GetStoredCStorVersionClient().CStorPoolInstances(cspiCopy.Namespace).Update(cspiCopy)
```
mittachaitu: Should we need to do updating CSP and patching deployment only in below condition
```go
if !reflect.DeepEqual(cspi.Spec.PoolConfig, cspiCopy.Spec.PoolConfig) {
// Step to perform update
// Step to perform patching of deployment
}
```

pkg/cspc/algorithm/build_deploy.go
```diff
@@ -401,5 +399,8 @@ func getPoolPodToleration(cspi *cstor.CStorPoolInstance) []corev1.Toleration {
 }
 
 func getPriorityClass(cspi *cstor.CStorPoolInstance) string {
-	return cspi.Spec.PoolConfig.PriorityClassName
+	if cspi.Spec.PoolConfig.PriorityClassName==nil{
```
mittachaitu: formating is required

pkg/controllers/cspc-controller/sync.go
```diff
@@ -261,3 +275,137 @@ func (c *Controller) GetCSPIWithoutDeployment(cspc *cstor.CStorPoolCluster) ([]c
...
+	if err != nil {
+		return errors.Wrap(err, "could not compute two way patch data")
+	}
+
+	strategicPatchData, err := strategicpatch.StrategicMergePatch(existingDeployObjInBytes, twoWayPatchData, v1.Deployment{})
+	if err != nil {
+		return errors.Wrap(err, "could not compute strategic patch data")
+	}
+
+	_, err = pc.Controller.kubeclientset.AppsV1().Deployments(cspi.Namespace).Patch(existingDeployObj.Name, k8stypes.StrategicMergePatchType, strategicPatchData)
```
mittachaitu: As discussed we need to handle an upgrade scenario where CSPC-Operator is updated but not CSPI we will take in next PR's.

PR #22, opened 2020-03-24
feat(operator): make openebs service account configurable for pool and volume deployments

prateekpandey14:
defaults to service account `openebs-maya-operator` in case
not provided.

refer: depends on https://github.com/openebs/api/pull/28

### DIFF
--- a/pkg/controllers/cstorvolumeclaim/deployment.go
+++ b/pkg/controllers/cstorvolumeclaim/deployment.go
@@ -67,9 +67,6 @@ var (
 			MountPath: "/var/openebs/cstor-target",
 		},
 	}
-	// OpenEBSServiceAccount name of the openebs service accout with required
-	// permissions
-	OpenEBSServiceAccount = "openebs-maya-operator"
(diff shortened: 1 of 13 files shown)


pkg/controllers/cstorvolumeclaim/deployment.go
```diff
@@ -201,6 +201,14 @@ func getTargetMgmtMounts() []corev1.VolumeMount {
 	return defaultMounts
 }
 
+func getServiceAccountName() string {
+	name, present := os.LookupEnv("OPENEBS_SERVICEACCOUNT_NAME")
```
mittachaitu: nit: We need to use some constants to represent these literals.

pkg/cspc/algorithm/build_deploy.go
```diff
@@ -255,6 +255,14 @@ func getMayaExporterImage() string {
 	return image
 }
 
+func getServiceAccountName() string {
```
mittachaitu: Can we move it to some common place? So that will be making use by all the operators that exist in this repo?

PR #26, opened 2020-03-30
feat(cspi): add capability to set compression on cstor pool

sonasingh46:
This commit adds capability to set compression on the cstor pool. The
Compression can be set via cspc and value must be one of 'on | off | lzjb |
gzip | gzip-[1-9] | zle | lz4'.

Note: lz4 is the defualt compression algorithm that is used if the compression field is left unspecified on the cspc.

### DIFF
--- a/pkg/controllers/cspi-controller/handler.go
+++ b/pkg/controllers/cspi-controller/handler.go
@@ -126,9 +126,9 @@ func (c *CStorPoolInstanceController) reconcile(key string) error {
 			string(common.SuccessCreated),
 			fmt.Sprintf("Pool created successfully"))
 
-		_, err := c.update(cspi)
+		cspiGot, err := c.update(cspi)
 		if err != nil {
-			c.recorder.Event(cspi,
+			c.recorder.Event(cspiGot,
 				corev1.EventTypeWarning,
 				string(common.FailedSynced),
 				err.Error())
@@ -191,7 +191,7 @@ func (c *CStorPoolInstanceController) update(cspi *cstor.CStorPoolInstance) (*cs
 		WithRecorder(c.recorder)
 	cspi, err := oc.Update(cspi)
 	if err != nil {
-		return nil, errors.Errorf("Failed to update pool due to %s", err.Error())
+		return cspi, errors.Errorf("Failed to update pool due to %s", err.Error())
 	}
 	return c.updateStatus(cspi)
 }
@@ -203,11 +203,11 @@ func (c *CStorPoolInstanceController) updateStatus(cspi *cstor.CStorPoolInstance
 	pool := zpool.PoolName()
 	propertyList := []string{"health", "io.openebs:readonly"}
 
-	// Since we quarried in following order health and io.openebs:readonly output also
+	// Since we queried in following order health and io.openebs:readonly output also
 	// will be in same order
 	valueList, err := zpool.GetListOfPropertyValues(pool, propertyList)
 	if err != nil {
-		return nil, errors.Errorf("Failed to fetch %v output: %v error: %v", propertyList, valueList, err)
+		return cspi, errors.Errorf("Failed to fetch %v output: %v error: %v", propertyList, valueList, err)
 	} else {
 		// valueList[0] will hold the value of health of cStor pool
 		// valueList[1] will hold the value of io.openebs:readonly of cStor pool
@@ -219,7 +219,7 @@ func (c *CStorPoolInstanceController) updateStatus(cspi *cstor.CStorPoolInstance
 
 	status.Capacity, err = zpool.GetCSPICapacity(pool)
 	if err != nil {
-		return nil, errors.Errorf("Failed to sync due to %s", err.Error())
+		return cspi, errors.Errorf("Failed to sync due to %s", err.Error())
 	}
 	c.updateROMode(&status, *cspi)
 
--- a/pkg/controllers/cspi-controller/handler.go
+++ b/pkg/controllers/cspi-controller/handler.go
@@ -89,13 +89,20 @@ func (c *CStorPoolInstanceController) reconcile(key string) error {
 		}
 		zpool.CheckImportedPoolVolume()
 		common.SyncResources.Mux.Unlock()
-		err = c.update(cspi)
+		cspiGot, err := c.update(cspi)
 		if err != nil {
-			c.recorder.Event(cspi,
+			c.recorder.Event(cspiGot,
 				corev1.EventTypeWarning,
 				string(common.FailedSynced),
 				err.Error())
 		}
+
+		// If everything is alright here -- sync the cspi
+		// Note: Even if update fails, cspiGot will not be nil.
+		// In case of failed update, passed cspi to update functions
+		// is returned.
+		c.sync(cspiGot)
+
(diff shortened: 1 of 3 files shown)


pkg/pool/operations/create.go
```diff
@@ -103,5 +105,16 @@ func (oc *OperationsConfig) createPool(cspi *cstor.CStorPoolInstance, r cstor.Ra
 	if err != nil {
 		return errors.Errorf("Failed to create pool.. %s .. %s", string(ret), err.Error())
 	}
+
+	// Set compression at the pool level
+	compressionType := cspi.Spec.PoolConfig.Compression
```
mittachaitu: Sorry for not giving this point before raising a PR. I verified quickly we can provide compression property during pool creation itself.
```go
ret, err := zfs.NewPoolCreate().
		WithType(ptype).
		WithProperty("cachefile", types.CStorPoolBasePath+types.CacheFileName).
                WithProperty("compression", cspi.Spec.PoolConfig.Compression).
		WithPool(PoolName()).
		WithVdevList(vdevlist).
		Execute()
```
But the value of `cspi.Spec.PoolConfig.Compression` shouldn't be empty.

pkg/pool/operations/pool_utils.go
```diff
@@ -32,6 +32,24 @@ import (
 	"k8s.io/klog"
 )
 
+var SupportedCompressionTypes = map[string]bool{
```
mittachaitu: nit: This kind of validation should be incorporated with openAPISchemaValidations :) 
https://github.com/openebs/zfs-localpv/pull/66 (But we will take it later).

pkg/controllers/cspi-controller/handler.go
```diff
@@ -346,3 +355,16 @@ func (c *CStorPoolInstanceController) addPoolProtectionFinalizer(
 		string(cspi.GetUID()))
 	return newCSPI, nil
 }
+
+func (c *CStorPoolInstanceController) sync(cspi *cstor.CStorPoolInstance) {
+	// Right now the only sync activity is compression
+	compressionType := cspi.Spec.PoolConfig.Compression
+	poolName := zpool.PoolName()
+	err := zpool.SetCompression(poolName, compressionType)
```
mittachaitu: Before setting we should do get if it doesn't match then good to set. Instead of setting for every reconciliation.

pkg/pool/operations/pool_utils.go
sonasingh46: I am aware of this. I have not yet started to work on OpenAPI CRD schema.
Even though we will have validations in OpenAPI schema — the code will have the critical validations in place.

pkg/controllers/cspi-controller/handler.go
sonasingh46: Not required IMO unless there are any specific reasons.

pkg/pool/operations/create.go
sonasingh46: Will do that.

pkg/controllers/cspi-controller/handler.go
mittachaitu: Here we are performing a write operation on the disk for every 30 sec which is not good.

pkg/controllers/cspi-controller/handler.go
```diff
@@ -119,7 +126,7 @@ func (c *CStorPoolInstanceController) reconcile(key string) error {
 			string(common.SuccessCreated),
 			fmt.Sprintf("Pool created successfully"))
 
-		err = c.update(cspi)
+		_, err := c.update(cspi)
```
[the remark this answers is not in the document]
sonasingh46: Not required but will do for consistency.

pkg/controllers/cspi-controller/handler.go
```diff
@@ -177,19 +184,19 @@ updatestatus:
...
-func (c *CStorPoolInstanceController) update(cspi *cstor.CStorPoolInstance) error {
+func (c *CStorPoolInstanceController) update(cspi *cstor.CStorPoolInstance) (*cstor.CStorPoolInstance, error) {
 	oc := zpool.NewOperationsConfig().
 		WithKubeClientSet(c.kubeclientset).
 		WithOpenEBSClient(c.clientset).
 		WithRecorder(c.recorder)
 	cspi, err := oc.Update(cspi)
 	if err != nil {
-		return errors.Errorf("Failed to update pool due to %s", err.Error())
+		return nil, errors.Errorf("Failed to update pool due to %s", err.Error())
```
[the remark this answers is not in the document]
sonasingh46: thanks

pkg/controllers/cspi-controller/handler.go
sonasingh46: Done.

pkg/pool/operations/pool_utils.go
```diff
@@ -295,3 +313,39 @@ func (oc *OperationsConfig) cleanUpReplacementMarks(oldObj, newObj *openebsapis.
 	}
 	return nil
 }
+
+// SetCompression sets the cstor pool compression
+func SetCompression(poolName string, compressionType string) error {
+	// If compression type is empty -- it means disable compression on the pool
```
pawanpraka1: let us act only when user has given something. Let us not do anything if compressionType is empty.

sonasingh46: Empty is an invalid value -- so in this case, will return an error.
Initially, I thought that empty means off so did like this.
Thanks.

pawanpraka1: what if user has not added compression in the CR, what you will get here?

mittachaitu: It will be empty what does it mean?

pawanpraka1: then we should not treat that as error. And we should not assume anything also. We should fire the command without the comression property.

sonasingh46: Makes sense. Will do that.

sonasingh46: Thanks Pawan

pkg/controllers/cspi-controller/handler.go
```diff
@@ -177,30 +184,30 @@ updatestatus:
...
-func (c *CStorPoolInstanceController) updateStatus(cspi *cstor.CStorPoolInstance) error {
+func (c *CStorPoolInstanceController) updateStatus(cspi *cstor.CStorPoolInstance) (*cstor.CStorPoolInstance, error) {
 	// ToDo: Use the status from the cspi object that is passed in arg else other fields
 	// might get lost.
 	var status cstor.CStorPoolInstanceStatus
 	pool := zpool.PoolName()
 	propertyList := []string{"health", "io.openebs:readonly"}
 
-	// Since we quarried in following order health and io.openebs:readonly output also
+	// Since we queried in following order health and io.openebs:readonly output also
```
mittachaitu: Thanks for fixing typo

pkg/pool/operations/pool_utils.go
sonasingh46: When I looked back, I think the current implementation looks good to me. 
Happy to discuss more.

PR #29, opened 2020-04-03
refact(cspc): add tests and constants for cspc conditions

sonasingh46:
This PR adds constants for cspc condition reason

### DIFF
--- a/pkg/controllers/cspc-controller/util/cspc_util_test.go
+++ b/pkg/controllers/cspc-controller/util/cspc_util_test.go
@@ -52,19 +52,15 @@ func TestGetCondition(t *testing.T) {
 		expected bool
 	}{
 		{
-			name: "condition exists",
-
+			name:     "condition exists",
 			status:   *exampleStatus,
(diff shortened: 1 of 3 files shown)


pkg/controllers/cspc-controller/util/cspc_util_test.go
```diff
@@ -0,0 +1,171 @@
...
+				t.Errorf("%s: expected condition to exist: %t, got: %t", test.name, test.expected, exists)
+			}
+		})
+	}
+}
+
+func TestSetCondition(t *testing.T) {
+	tests := []struct {
+		name string
+
```
prateekpandey14, line 86: remove extra lines , i think go fmt run required in all these files .

sonasingh46: Go fmt does not remove extra lines. I think I need to remove it.

prateekpandey14: thanks

PR #30, opened 2020-04-04
chore(cspc): refactor cspc reconcile and cleanup

sonasingh46:
This PR does following enhancements:

1. If there is a mismatch b/w cspc and cspc-operator version the cspc
   is not reconciled. If the cspc has an empty version then cspc is reconiled.

2. The cspc and cspi cleanup function is improved and test cases are added
   with fakecontrollers.

### DIFF
--- a/pkg/controllers/cspc-controller/sync.go
+++ b/pkg/controllers/cspc-controller/sync.go
@@ -36,12 +36,26 @@ import (
 )
 
 func (c *Controller) sync(cspc *cstor.CStorPoolCluster, cspiList *cstor.CStorPoolInstanceList) error {
+	if ok, reason := c.ShouldReconcile(*cspc); !ok {
+		// Do not reconcile this cspc
+		message := fmt.Sprintf("Cannot not reconcile CSPC %s as %s", cspc.Name, reason)
+		c.recorder.Event(cspc, corev1.EventTypeWarning, "CSPC Reconcile", message)
+		klog.Warningf("Cannot not reconcile CSPC %s in namespace %s as %s", cspc.Name, cspc.Namespace, reason)
+		return nil
+	}
 
-	// If CSPC is deleted -- delete all the associated CSPI resources.
-	// Cleaning up CSPI resources in case of removing poolSpec from CSPC
+	// cleaning up CSPI resources in case of removing poolSpec from CSPC
 	// or manual CSPI deletion
 	if cspc.DeletionTimestamp.IsZero() {
-		err := c.cleanupCSPIResources(cspc)
+		cspiList, err := c.GetCSPIListForCSPC(cspc)
+		if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/controllers/cspc-controller/sync.go
```diff
@@ -438,3 +458,17 @@ func (pc *PoolConfig) patchPoolDeploymentSpec(cspi *cstor.CStorPoolInstance) err
 
 	return nil
 }
+
+func (c *Controller) ShouldReconcile(cspc cstor.CStorPoolCluster) (bool, string) {
+	cspcOperatorVersion := version.Current()
```
mittachaitu, line 463: We will maintain Version file in cstor-opeeraters repo instead of depending on Maya

pkg/controllers/cspc-controller/sync.go
```diff
@@ -438,3 +458,17 @@ func (pc *PoolConfig) patchPoolDeploymentSpec(cspi *cstor.CStorPoolInstance) err
 
 	return nil
 }
+
+func (c *Controller) ShouldReconcile(cspc cstor.CStorPoolCluster) (bool, string) {
```
mittachaitu, line 462: Why return type string?Can we have a return type error?

sonasingh46: As discussed -- we want to have the reason of truthy value and not error occurs here.

pkg/controllers/cspc-controller/sync.go
sonasingh46: Will migrate version package to cstor-operators

pkg/controllers/cspc-controller/sync.go
```diff
@@ -67,7 +81,9 @@ func (c *Controller) sync(cspc *cstor.CStorPoolCluster, cspiList *cstor.CStorPoo
 	if !cspcGot.DeletionTimestamp.IsZero() {
 		err = c.handleCSPCDeletion(cspcGot)
 		if err != nil {
-			klog.Errorf("Failed to sync CSPC %s in namespace %s for deletion:%s", cspc.Name, cspc.Namespace, err.Error())
+			message := fmt.Sprintf("Could not sync for CSPC:{%s} deletion", err.Error())
+			c.recorder.Event(cspc, corev1.EventTypeWarning, "CSPC Cleanup", message)
+			klog.Errorf("Failed to cleanup CSPC %s in namespace %s: %s", cspc.Name, cspc.Namespace, err.Error())
```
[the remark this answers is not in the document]
sonasingh46: The recorder is not logging in the cspc-operator.

pkg/controllers/cspc-controller/sync.go
sonasingh46: @mittachaitu -- migration of version package will be taken in a separate PR.

PR #43, opened 2020-04-21
feat(pool expansion): add a support to perform pool expansion when underlying disk expanded

mittachaitu:

This PR adds support to perform pool expansion when underlying disk expanded.

**How Users/Admins Can Make Use Of It**:
- User/Admin has to make sure that expanded disk capacity should be reflected in corresponding Blocckdevice CR(As of one option is by restarting the NDM daemonset pod scheduled on that machine).

**Success Events**:
CMD: `kubectl describe cspi <cspi_name> -n <openebs_namespace>`
```sh
Events:
  Type     Reason                       Age                   From               Message
  ----     ------                       ----                  ----               -------
  Normal   Created                      6m56s                 CStorPoolInstance  Pool created successfully
  Normal   Pool Expansion               56s                   CStorPoolInstance  Expanded the disks which were in use by pool
```

This is the development PR of OEP: https://github.com/openebs/openebs/pull/3001
It fixes the issue: https://github.com/openebs/openebs/issues/2372


pkg/pool/operations/modify.go
```diff
@@ -191,5 +196,57 @@ func (oc *OperationsConfig) Update(cspi *cstor.CStorPoolInstance) (*cstor.CStorP
...
+// ExpandPoolIfDiskExpanded performs pool expansion when underlying disk got expanded
+// currently it will identify by comparing the capacity of blockdevice and capacity exist
+// on CStorPoolInstanceBlockDevice
+func (oc *OperationsConfig) ExpandPoolIfDiskExpanded(
+	cspi *cstor.CStorPoolInstance) (*cstor.CStorPoolInstance, error) {
+	var err error
+	var isPoolExpanded bool
+	minimumMB := uint64(500 * 1024 * 1024)
+	openebsNamespace := os.Getenv(string(common.OpenEBSIOPoolName))
+	for _, raidGroup := range cspi.Spec.DataRaidGroups {
```
mittachaitu, line 218: TODO: Need to check can we able to expand write-cache raid groups

pkg/partition/partprobe/builder.go
```diff
@@ -0,0 +1,90 @@
...
+// Execute is to execute generated partprobe against disk object
+func (d *Disk) Execute() ([]byte, error) {
+	d, err := d.Build()
+	if err != nil {
+		return nil, err
+	}
+	fmt.Printf("Command %s\n", d.Command)
+	// execute command here
+	// #nosec
+	return exec.Command("bash", "-c", d.Command).CombinedOutput()
```
akhilerm, line 64: Why is the command executed in a bash shell? Can't it be run directly?

mittachaitu: In the CSPI controller we are following this practice executing in bash shell so followed the same practice.

pkg/partition/partprobe/builder.go
```diff
@@ -0,0 +1,90 @@
+/*
```
sonasingh46, line 1: n_p: moving this to package pkg/pool/partition can help organize a little better. 
What are your thoughts?

pkg/partition/partprobe/builder.go
```diff
@@ -0,0 +1,90 @@
...
+type Disk struct {
+	// Device represents the path to disks
+	Device []string
+
+	// error
+	err error
+
+	// command string
+	Command string
+}
```
sonasingh46, line 42: n_p: these extra spaces can be removed.

pkg/partition/partprobe/builder.go
sonasingh46: Also, file name builder.go becomes too generic IMO.

sonasingh46: `partprobe.go` sounds good to me and the `predicate.go` content can move here. 
Idea is to avoid getting into imparting the creational patterns.

pkg/partition/partprobe/builder.go
```diff
@@ -0,0 +1,90 @@
...
+
+// NewDisk returns new instance of object Disk
+func NewDisk() *Disk {
+	return &Disk{}
+}
+
+// WithDevice method fills the Device field of PoolOnline object.
```
sonasingh46, line 49: What is a PoolOnline object?

pkg/partition/partprobe/builder.go
```diff
@@ -0,0 +1,90 @@
...
+	// Operation represents the command that executes against provided devices
+	Operation = "partprobe"
+)
+
+// Disk represents the for 'partprobe' operation
+type Disk struct {
+	// Device represents the path to disks
+	Device []string
+
+	// error
```
sonasingh46, line 37: this comment is not required.

pkg/partition/partprobe/builder.go
```diff
@@ -0,0 +1,90 @@
...
+	err error
+
+	// command string
```
sonasingh46, line 40: this comment can be improved.

pkg/partition/partprobe/builder.go
```diff
@@ -0,0 +1,90 @@
...
+	"fmt"
+	"os/exec"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+const (
+	// Operation represents the command that executes against provided devices
+	Operation = "partprobe"
```
sonasingh46, line 29: If this is a command -- then
```
PartProbeCommand= "partprobe"
```
Because Operation cannot represent anything else

pkg/partition/partprobe/builder.go
```diff
@@ -0,0 +1,90 @@
...
+	return d
+}
+
+// Execute is to execute generated partprobe against disk object
+func (d *Disk) Execute() ([]byte, error) {
+	d, err := d.Build()
+	if err != nil {
+		return nil, err
+	}
+	fmt.Printf("Command %s\n", d.Command)
```
sonasingh46, line 61: Why printf ? Any specific reason? 
Can't we use logger?

pkg/partition/partprobe/builder.go
mittachaitu: Sounds good

pkg/partition/partprobe/builder.go
mittachaitu: It should be Disk object I will update.

pkg/partition/partprobe/builder.go
mittachaitu: Not required I will remove it.

mittachaitu:
Closing this PR since it is not actively worked and it can be re-opened based on community request.

PR #46, opened 2020-04-27
feat(webhook): add cspc expansion validations

shubham14bajpai:

This PR adds webhook validations for cspc expansion.

Refer: https://github.com/openebs/maya/pull/1588

TODO:
add unit test for new functions

### DIFF
--- a/pkg/webhook/cspc_operations.go
+++ b/pkg/webhook/cspc_operations.go
@@ -32,6 +32,11 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+const (
+	dataRG       = "data"
+	writeCacheRG = "writeCache"
+)
+
 // PoolOperations contains old and new CSPC to validate for pool
 // operations
 type PoolOperations struct {
@@ -554,17 +559,17 @@ func (pOps *PoolOperations) validatePoolExpansion(
 			bds = append(bds, getNewBDsFromStripeSpec(rgs.oldRaidGroups[0],
 				rgs.newRaidGroups[0])...)
 		} else {
-			if rgType == "data" {
+			if rgType == dataRG {
 				newRgs := getExpandedRaidGroups(newPoolSpec.DataRaidGroups, rgs.oldRaidGroups)
 				bds = getBDsFromRaidGroups(newRgs)
 			}
-			if rgType == "writeCache" {
+			if rgType == writeCacheRG {
 				newRgs := getExpandedRaidGroups(newPoolSpec.WriteCacheRaidGroups, rgs.oldRaidGroups)
 				bds = getBDsFromRaidGroups(newRgs)
 			}
 		}
 	}
-	err := pOps.validateNewBDs(bds, pOps.NewCSPC)
+	err := pOps.validateNewBDs(bds, pOps.OldCSPC)
 	if err != nil {
 		return err
 	}
@@ -577,12 +582,12 @@ func (pOps *PoolOperations) validatePoolExpansion(
 func getIndexedCommonRaidGroups(oldPoolSpec,
 	newPoolSpec *cstor.PoolSpec) (map[string]*raidGroups, error) {
 	rgs := map[string]*raidGroups{
-		"data": &raidGroups{
+		dataRG: &raidGroups{
 			oldRaidGroups: []cstor.RaidGroup{},
 			newRaidGroups: []cstor.RaidGroup{},
 			rgType:        oldPoolSpec.PoolConfig.DataRaidGroupType,
 		},
-		"writeCache": &raidGroups{
+		writeCacheRG: &raidGroups{
 			oldRaidGroups: []cstor.RaidGroup{},
 			newRaidGroups: []cstor.RaidGroup{},
 			rgType:        oldPoolSpec.PoolConfig.WriteCacheGroupType,
@@ -594,8 +599,8 @@ func getIndexedCommonRaidGroups(oldPoolSpec,
 		for _, newRg := range newPoolSpec.DataRaidGroups {
 			if IsRaidGroupCommon(oldRg, newRg) {
 				isRaidGroupExist = true
-				rgs["data"].oldRaidGroups = append(rgs["data"].oldRaidGroups, oldRg)
-				rgs["data"].newRaidGroups = append(rgs["data"].newRaidGroups, newRg)
+				rgs[dataRG].oldRaidGroups = append(rgs[dataRG].oldRaidGroups, oldRg)
+				rgs[dataRG].newRaidGroups = append(rgs[dataRG].newRaidGroups, newRg)
 				break
 			}
 		}
@@ -609,8 +614,8 @@ func getIndexedCommonRaidGroups(oldPoolSpec,
 		for _, newRg := range newPoolSpec.WriteCacheRaidGroups {
 			if IsRaidGroupCommon(oldRg, newRg) {
 				isRaidGroupExist = true
-				rgs["writeCache"].oldRaidGroups = append(rgs["writeCache"].oldRaidGroups, oldRg)
-				rgs["writeCache"].newRaidGroups = append(rgs["writeCache"].newRaidGroups, newRg)
+				rgs[writeCacheRG].oldRaidGroups = append(rgs[writeCacheRG].oldRaidGroups, oldRg)
+				rgs[writeCacheRG].newRaidGroups = append(rgs[writeCacheRG].newRaidGroups, newRg)
 				break
 			}
 		}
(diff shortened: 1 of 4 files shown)


pkg/webhook/cspc_operations.go
```diff
@@ -509,3 +468,176 @@ func GetNewBDFromRaidGroups(newRG, oldRG *cstor.RaidGroup) map[string]string {
...
+// validateNewBDs returns nil if new BDs are valid for expansion or else it will
+// return error if occurs
+func (pOps *PoolOperations) validateNewBDs(newBDs []string, cspc *cstor.CStorPoolCluster) error {
+	for _, bd := range newBDs {
+		bdc, err := pOps.GetBDCOfBD(bd)
+		if err != nil {
+			return errors.Wrapf(err, "failed to get claim of block device %s", bd)
+		}
+		// New BD should be valid only if:
+		// 1. If BlockDevice already claimed that claim should have current CSPC mark.
```
mittachaitu, line 534: nit: Update comments to
```
	// The incoming BD should be a valid entry if
	// 1. The BD does not have a BDC.
	// 2. The BD has a BDC with the current CSPC label and there is no successor of this BD
	//    present in the CSPC.
```

pkg/webhook/cspc_operations.go
```diff
@@ -509,3 +470,203 @@ func GetNewBDFromRaidGroups(newRG, oldRG *cstor.RaidGroup) map[string]string {
...
+
+// validatePoolExpansion will validate only expanded raid groups or new blockdevices(
+// in stripe only block devices are added). Following are the validations:
+// 1. New blockdevice shouldn't be claimed by any other CSPC (or) third party.
+// 2. New blockdevice shouldn't be the replacing blockdevice.
+func (pOps *PoolOperations) validatePoolExpansion(
+	newPoolSpec *cstor.PoolSpec, commonRaidGroups *raidGroups, rgType string) error {
+	var bds []string
+	// Multiple raidGroups doesn't exist for stripe pool spec
+	if _, ok := SupportedPRaidType[cstor.PoolType(rgType)]; ok {
```
mittachaitu, line 555: This check should be against the stripe pool. Since in stripe, only one raid group is allowed

pkg/webhook/cspc_operations.go
```diff
@@ -509,3 +470,203 @@ func GetNewBDFromRaidGroups(newRG, oldRG *cstor.RaidGroup) map[string]string {
...
+					for k, v := range newBD {
+						newToOldBd[k] = v
+					}
+				}
+				// Already mapped(via index) old raid groups and new raid groups in
+				// commonRaidGroups no need to check other raid groups
+				break
+			}
+		}
+		err = pOps.validatePoolExpansion(newPoolSpec, v, rgType)
```
mittachaitu, line 659: It should be out of the loop and pass the two arguments newPoolSpec & commonRaidGroups. Then we have to change the function also.
If we write out of the loop then we will not check again and again please look at func definition comment

pkg/webhook/cspc_operations.go
```diff
@@ -509,3 +470,203 @@ func GetNewBDFromRaidGroups(newRG, oldRG *cstor.RaidGroup) map[string]string {
...
+
+// raidGroups contains list of oldraid groups and newraid groups
+type raidGroups struct {
+	oldRaidGroups []cstor.RaidGroup
+	newRaidGroups []cstor.RaidGroup
+	rgType        string
+}
+
+func getNewBDsFromStripeSpec(oldRg, newRg cstor.RaidGroup) []string {
+	mapOldRg := map[string]bool{}
```
mittachaitu, line 482: nit: good to change the variable name to mapOldBlockDevices so that it will be meaningful

pkg/webhook/cspc_operations.go
```diff
@@ -509,3 +470,203 @@ func GetNewBDFromRaidGroups(newRG, oldRG *cstor.RaidGroup) map[string]string {
...
+		}
+	}
+	return nil
+}
+
+// validatePoolExpansion will validate only expanded raid groups or new blockdevices(
+// in stripe only block devices are added). Following are the validations:
+// 1. New blockdevice shouldn't be claimed by any other CSPC (or) third party.
+// 2. New blockdevice shouldn't be the replacing blockdevice.
+func (pOps *PoolOperations) validatePoolExpansion(
```
mittachaitu, line 551:
```go
func (pOps *PoolOperations) validatePoolExpansion(
newPoolSpec *cstor.PoolSpec, commonRaidGroups map[string]*raidGroups) {
var bds []string 
for _, rgs := range commonRaidGroups {
   if rgs.rgType == "stripe" {
       bds = append(bds, getNewBDsFromStripeSpec(rgs.oldRaidGroups[0],
			rgs.newRaidGroups[0]))
   } else {
          newRgs := getExpandedRaidGroups(newPoolSpec, rgs.oldRaidGroups)
          bds = getBDsFromRaidGroups(newRgs)
      }
}
  err := pOps.validateNewBDs(bds, pOps.NewCSPC)
	if err != nil {
		return err
	}
	return nil
}
```
Check above will fit into it by considering caller code comment also

pkg/webhook/cspc_operations.go
```diff
@@ -509,3 +470,203 @@ func GetNewBDFromRaidGroups(newRG, oldRG *cstor.RaidGroup) map[string]string {
...
+		for _, bd := range rg.CStorPoolInstanceBlockDevices {
+			bds = append(bds, bd.BlockDeviceName)
+		}
+	}
+	return bds
+}
+
+// getExpandedRaidGroups returns the raid groups only if pool expansion is done
+// by the users
+func getExpandedRaidGroups(
```
mittachaitu, line 508: This function logic should be changed according  we have check-in writecache group also

pkg/webhook/cspc_operations.go
```diff
@@ -509,3 +468,213 @@ func GetNewBDFromRaidGroups(newRG, oldRG *cstor.RaidGroup) map[string]string {
...
+			rgType:        oldPoolSpec.PoolConfig.WriteCacheGroupType,
+		},
+	}
+	// build raidGroups by identifying common raidGroups in old and new
+	for _, oldRg := range oldPoolSpec.DataRaidGroups {
+		isRaidGroupExist := false
+		for _, newRg := range newPoolSpec.DataRaidGroups {
+			if IsRaidGroupCommon(oldRg, newRg) {
+				isRaidGroupExist = true
+				rgs["data"].oldRaidGroups = append(rgs["data"].oldRaidGroups, oldRg)
```
mittachaitu, line 597: nit: IMO good to maintain global variable for the type of the disk instead of literals

pkg/webhook/cspc_operations.go
```diff
@@ -282,9 +237,9 @@ func (bdr *BlockDeviceReplacement) IsBDReplacementValid(newRG, oldRG *cstor.Raid
 
 // validateNewBDCapacity returns error only when new block device has less capacity
 // than existing block device
-func (bdr *BlockDeviceReplacement) validateNewBDCapacity(newRG, oldRG *cstor.RaidGroup) error {
+func (pOps *PoolOperations) validateNewBDCapacity(newRG, oldRG *cstor.RaidGroup) error {
```
mittachaitu, line 240: TODO: For me, in next PR's we can call `validateNewBDCapacity` function in `AreNewBDsValid` so that function name will full fill the need.

pkg/webhook/cspc_operations.go
```diff
@@ -509,3 +468,213 @@ func GetNewBDFromRaidGroups(newRG, oldRG *cstor.RaidGroup) map[string]string {
...
+				newRgs := getExpandedRaidGroups(newPoolSpec.DataRaidGroups, rgs.oldRaidGroups)
+				bds = getBDsFromRaidGroups(newRgs)
+			}
+			if rgType == "writeCache" {
+				newRgs := getExpandedRaidGroups(newPoolSpec.WriteCacheRaidGroups, rgs.oldRaidGroups)
+				bds = getBDsFromRaidGroups(newRgs)
+			}
+		}
+	}
+	err := pOps.validateNewBDs(bds, pOps.NewCSPC)
```
mittachaitu, line 567: In validateNewBDs function we are calling `IsBDValid`. IsBDValid function is expecting oldCSPC but we passed NewCSPC. If we look from here it makes sense to pass oldCSPC since we already have new blockDevices in our hand.
```go
func (pOps *PoolOperations) IsBDValid(bd string, bdc *openebsapis.BlockDeviceClaim, oldcspc *cstor.CStorPoolCluster) bool {
}
```

pkg/webhook/cspc_operations.go
```diff
@@ -327,41 +285,43 @@ func (bdr *BlockDeviceReplacement) IsNewBDPresentOnCurrentCSPC(newRG, oldRG *cst
...
 	for _, v := range oldRG.CStorPoolInstanceBlockDevices {
-		bdcObject, err := bdr.GetBDCOfBD(v.BlockDeviceName)
+		bdcObject, err := pOps.GetBDCOfBD(v.BlockDeviceName)
 		if err != nil {
 			return true, errors.Errorf("failed to query for any existing replacement in the raid group : %s", err.Error())
 		}
-		_, ok := bdcObject.GetAnnotations()[types.PredecessorBDLabelKey]
-		if ok {
-			return true, errors.Errorf("replacement is still in progress for bd %s", v.BlockDeviceName)
+		if bdcObject != nil {
```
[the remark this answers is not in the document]
sonasingh46: This can be improved by returning a typed error -- like BDC does not exists. 
And BDC does not exist --  should be done instead. 
Philosophy is -- if err is nil -- object should be not nil -- code looks more cleaner.
Commenting for tracking. It can be enhanced in a new PR.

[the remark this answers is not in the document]
sonasingh46: Anyways -- thanks. Nice Catch.

PR #48, opened 2020-04-28
fix(deps): remove indirect version dependency to openebs/api

prateekpandey14:
Refer : https://github.com/openebs/api/pull/43


pkg/version/util.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"strings"
+)
+
+var (
+	validCurrentVersions = map[string]bool{
+		"1.9.0": true, "1.10.0": true,
+	}
+	validDesiredVersion = GetVersion()
```
mittachaitu, line 11: https://github.com/openebs/cstor-operators/blob/853f02b673948170779407bce1e2f269b0e5f555/pkg/version/version.go#L41
We need to maintain one VERSION file in cstor-operaters repo.

prateekpandey14: yes , the file path has been updated

PR #60, opened 2020-05-07
feat(upgrade): add upgrade framework in controllers

shubham14bajpai:

This PR adds an upgrade framework for all the controllers to perform any upgrade steps from within the controller itself to avoid any pod exec issues.

### DIFF
--- a/pkg/controllers/cspc-controller/sync.go
+++ b/pkg/controllers/cspc-controller/sync.go
@@ -51,8 +51,21 @@ var (
 
 func (c *Controller) sync(cspc *cstor.CStorPoolCluster, cspiList *cstor.CStorPoolInstanceList) error {
 
+	// If deletion timestamp is not zero on CSPC, this means CSPC is deleted
+	// and all the resources associated with cspc should be deleted.
+	if !cspc.DeletionTimestamp.IsZero() {
+		err := c.handleCSPCDeletion(cspc)
+		if err != nil {
+			message := fmt.Sprintf("Could not sync for CSPC:{%s} deletion", err.Error())
+			c.recorder.Event(cspc, corev1.EventTypeWarning, "CSPC Cleanup", message)
+			klog.Errorf("Failed to cleanup CSPC %s in namespace %s: %s", cspc.Name, cspc.Namespace, err.Error())
+		}
+		return nil
+	}
+
 	// cleaning up CSPI resources in case of removing poolSpec from CSPC
 	// or manual CSPI deletion
+	// This should be performed before reconcileVersion is done.
 	if cspc.DeletionTimestamp.IsZero() {
 		cspiList, err := c.GetCSPIListForCSPC(cspc)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/controllers/cspc-controller/sync.go
```diff
@@ -36,6 +38,17 @@ import (
 	"k8s.io/klog"
 )
 
+type upgradeParams struct {
+	cspc   *cstor.CStorPoolCluster
+	client clientset.Interface
+}
+
+type upgradeFunc func(u *upgradeParams) (*cstor.CStorPoolCluster, error)
```
mittachaitu, line 46: Will this make etcd calls to update CSPC? If yes just thinking why can't we make this internal Upgrade as an atomic operation? Is there any reason why we not selected?

pkg/controllers/cstorvolumeconfig/controller.go
```diff
@@ -156,6 +169,25 @@ func (c *CVCController) syncCVC(cvc *apis.CStorVolumeConfig) error {
 	}
 
 	var err error
+
+	cvc, err = c.reconcileVersion(cvc)
```
mittachaitu, line 173: Before calling reconcileVersion, we need to populate CVC version details.

shubham14bajpai:
> Can we use these functions not only to perform exec but also for any schema changes i.e adding field? Instead of prepopulating by upgrade job?

Yes we can edit the fields according to requirements in these functions also.

pkg/controllers/cspc-controller/sync.go
shubham14bajpai: Some times the operations may not be atomic or we have steps that need to be performed in some order. If by atomic you mean single operation then it is safer to have a framework that allows you to have both options of having separate or a single operation. We can always write functions that don't make any etcd calls if not required but having the option will be better for any future requirements.

pkg/controllers/cstorvolumeconfig/controller.go
shubham14bajpai: Version will be populated while creation itself for CVC.

pkg/controllers/cspc-controller/sync.go
```diff
@@ -71,6 +77,32 @@ func (c *Controller) sync(cspc *cstor.CStorPoolCluster, cspiList *cstor.CStorPoo
...
+			err,
+		)
+		_, err = c.clientset.CstorV1().CStorPoolClusters(cspcGot.Namespace).Update(cspcGot)
+		if err != nil {
+			klog.Errorf("failed to update versionDetails status for cspc %s:%s", cspcGot.Name, err.Error())
+		}
+		return nil
+	}
+
+	if ok, reason := c.ShouldReconcile(*cspc); !ok {
```
mittachaitu, line 98: It should move below the code
```go
	if !cspcGot.DeletionTimestamp.IsZero() {
}
```

shubham14bajpai: When the controller is upgraded then reconcile version will never be triggered after patching the VersionDetails.DesiredVersion to the upgraded version as the ShouldReconcile function checks for CurrentVersion. Once the reconcileVersion is successful the CurrentVersion on the CR will be changed to the new version.

mittachaitu: I got it but we can delete the resource irrespective of current version so need to move below https://github.com/openebs/cstor-operators/blob/bc2114c8698a27a602fb55f419618de7e9fe16d4/pkg/controllers/cspc-controller/sync.go#L82

(Or)

[This snippet](https://github.com/openebs/cstor-operators/blob/bc2114c8698a27a602fb55f419618de7e9fe16d4/pkg/controllers/cspc-controller/sync.go#L82) should move before any function.

shubham14bajpai: Got it! Nice Catch :+1:

PR #66, opened 2020-05-11
chore(ci): fast fail travis script in case of error

sonasingh46:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ before_script: # TODO add golangci yaml config
 script:
   - make test
   - if [ "$TRAVIS_CPU_ARCH" == "amd64" ]; then
-      make all.amd64;
+      make all.amd64 || exit 1;
     fi
   # Install openebs-operators, cstor-csi, ndm-operators and other
   # dependencies.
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -51,7 +51,7 @@ before_script: # TODO add golangci yaml config
 script:
   - make test
   - if [ "$TRAVIS_CPU_ARCH" == "amd64" ]; then
-      make all.amd64;
+      make all.amd64 || exit 1;
```
prateekpandey14, line 54: we can use `travis_terminate 1;`

sonasingh46: No documentation found regarding that. 
Found the `travis_terminate 1` thing in a deprecated api.
I asked the same in the Travis community.
https://travis-ci.community/t/how-to-stop-the-execution-of-travis-if-script-exits-with-an-error/8556

ci/sanity/install.sh
```diff
@@ -1,4 +1,7 @@
 #!/usr/bin/env bash
+
```
kmova, line 2: though not related to current PR, whenever we are touching any existing files, if a copyright message is missing, let us add one.

.travis.yml
```diff
@@ -1,3 +1,17 @@
+# Copyright © 2020 The OpenEBS Authors
```
kmova, line 1: :+1:

PR #68, opened 2020-05-12
chore(log): add cspc information in error event

mittachaitu:
This PR adds the CPSC information in the error log. When the user Provision a volume with storage replciaCount > no.of available pools (or) when the user mentioned a typo of CSPC name then event will be shown.

### DIFF
--- a/pkg/controllers/cstorvolumeconfig/volume_operations.go
+++ b/pkg/controllers/cstorvolumeconfig/volume_operations.go
@@ -361,7 +361,12 @@ func (c *CVCController) distributeCVRs(
 	}
 
 	if len(usablePoolList.Items) < pendingReplicaCount {
-		return errors.New("not enough pools available to create replicas")
+		return errors.Errorf(
+			"not enough pools are available of provided CSPC: %q, usable pool count: %d pending replica count: %d",
+			cspcName,
+			len(usablePoolList.Items),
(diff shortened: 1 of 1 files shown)


pkg/controllers/cstorvolumeconfig/volume_operations.go
```diff
@@ -361,7 +361,12 @@ func (c *CVCController) distributeCVRs(
 	}
 
 	if len(usablePoolList.Items) < pendingReplicaCount {
-		return errors.New("not enough pools available to create replicas")
+		return errors.Errorf(
+			"not enough pools available are available of CSPC: %s usable pool count %d to create %d replicas",
```
prateekpandey14, line 365: available repeated ?
change to  --> "not enough pools are available of provided CSPC: " ?

mittachaitu: How about this?
```go
"not enough pools available are available of CSPC: %s usable pool count %d to create %d replicas"
```

jadsy2107: > available repeated ? change to --> "not enough pools are available of provided CSPC: " ?

I'm getting this exact error, the only place ive found on github and google with these words is here, i cant find a fix! 
Any thoughts ?

Issue #74, opened 2020-05-15
cspi status should show usable pool capacity


mittachaitu:
As of know `kubectl get cspi -n openebs` shows sum of all the raid groups in the pool but for usability purposes, it needs to show the usable capacity of the pool after excluding metadata and raid parity.

PR #75, opened 2020-05-18
fix(cspi): update the Capacity fields of CStorPoolInstance

mittachaitu:

This PR updates the fields of CStorPoolInstanceCapacity:
**From**
```go
type CStorPoolInstanceCapacity struct {
	Total resource.Quantity `json:"total"`
	Free  resource.Quantity `json:"free"`
	Used  resource.Quantity `json:"used"`
}
```
**To**
```go
// CStorPoolInstanceCapacity stores the pool capacity related attributes.
type CStorPoolInstanceCapacity struct {
	Used resource.Quantity `json:"used"`
	Free resource.Quantity `json:"free"`
	Total resource.Quantity `json:"total"`
	ZFS ZFSCapacityAttributes `json:"zfs"`
}
type ZFSCapacityAttributes struct {
	LogicalUsed resource.Quantity `json:"logicalUsed"`
}
```
Til now capacity information is fetched using `**zpool**` utility.
Since zpool utility doesn't exclude the metadata and raid parities
of the pool during the capacity report so, it shows the Total Capacity
of the pool is a sum of all the Data RaidGroups because of this
there might be few difficulties in understanding pool usage.
To fix this problem we are fetching the capacity details using **zfs utility** and
fields also updated related to fields reported by ZFS.

**Mapping b/w CustomColumns and ZFS comamds**:

- Output of `zfs get used <pool_dataset_name>` is mapped to `CSPI.Status.Capacity.Used` and the corresponding custom column is reported as **Allocated**.
- Output of `zfs get available <pool_dataset_name>` is mapped to 
`CSPI.Status.Capacity.Free` and the corresponding custom column is reported as **Free**.
- Sum of `CSPI.Status.Capacity.Used+CSPI.Status.Capacity.Free` is mapped to `CSPI.Status.Capacity.Total`and the corresponding custom column is reported as **Capacity**.

**Sample Output**:
```sh
sai@sai:~/Desktop/learn/aws_external_disk/experiment/v1test$ kubectl get cspi -n openebs
NAME               HOSTNAME    ALLOCATED   FREE    CAPACITY   READONLY   TYPE     STATUS   AGE
cspc-stripe-6l27   127.0.0.1   53k         9630M   9630053k   false      mirror   ONLINE   36s
```

**Special Note To Reviewers**:
- As part of this changes percentage calculation of pool usage has been modified with the following formulae: **(Used/ (Free+Used))\*100** based on this calculation if pool usage is >= roThresholdLimit then the pool is updated as ReadOnly mode.
- Explanation of fields available in PR description https://github.com/openebs/api/pull/47.

This PR fixes the issue: https://github.com/openebs/cstor-operators/issues/74


pkg/pool/operations/status.go
```diff
@@ -64,6 +65,26 @@ func GetListOfPropertyValues(poolName string, propertyList []string) ([]string,
 
 }
 
+// GetVolumeListPropertyValues will return value list for given property list
+// NOTE: It will return the property values in the same order as property list
+func GetVolumeListPropertyValues(dataSetName string, propertyList []string) ([]string, error) {
```
mynktl, line 70: Better to move it in `pkg/volumereplica/volumereplica.go`

pkg/pool/operations/status.go
```diff
@@ -44,9 +45,9 @@ func GetPropertyValue(poolName, property string) (string, error) {
 	return outStr[0], nil
 }
 
-// GetListOfPropertyValues will return value list for given property list
+// GetPoolListPropertyValues will return value list for given property list
 // NOTE: It will return the property values in the same order as property list
-func GetListOfPropertyValues(poolName string, propertyList []string) ([]string, error) {
+func GetPoolListPropertyValues(poolName string, propertyList []string) ([]string, error) {
```
mynktl, line 50: nit: since it is under pkg/pool/operation, better to not have pool phrase in function name.

avishnu:
@mittachaitu Have we removed the Total Capacity field?
Total resource.Quantity `json:"total"`
Can we add it back, will add more meaning to "kubectl get cspi" command...

avishnu:
Also, want to understand what is the difference between "used" and "logicalused" for a pool ? From the capacity monitoring perspective, which one should be considered?

mittachaitu:
> Also, want to understand what is the difference between "used" and "logicalused" for a pool ? From the capacity monitoring perspective, which one should be considered?

IMO if we want to monitor pool usage then good to monitor via `cspi.Status.Used`(logicalused is something similar to all the application data that we no need to worry) because it will give the space consumed on physical devices.

prateekpandey14:
lets create an issue for representing the bytes to a readable format ( Mi, Gi, Ti etc) to track this further, can be enhanced in upcoming PRs

PR #77, opened 2020-05-19
fix(cspi): import pool without cachefile

mittachaitu:
This PR will try to import the pool without the cache file as a last option. 
It will import the pool in the following ways:
- Import the pool using cachefile  `zpool import -c cachefile=<cache_file> -o cachefile=<cache_file> <pool_name>`
- Import the pool without using cachefile:
  1. If it is sparse based pool then command will be `zpool import  -o cachefile=<cache_file> -d <parent_directory_of_sparse_file> <pool_name>`
  2. If it device based pool then command will be `zpool import -o cachefile=<cache_file> <pool_name>`(By default it will scan /dev directory).

**NOTE**:
- Due to bug in persisting the cachefile i.e cachefile is not stored in the persistent path because of this when pool pod got restarted then the pool will not be imported to fix this added the code to import without cachefile.

### DIFF
--- a/ci/sanity/sanity.sh
+++ b/ci/sanity/sanity.sh
@@ -136,7 +136,7 @@ if [ $? -ne 0 ]; then
     exit 1
 fi
 ## verify whether cache file present in persistent path
-ls -lrth /var/openebs/cstor-pool/${cspc_name}/pool.cache
+ls -lrth /var/openebs/cstor-pool/${cspcName}/pool.cache
 if [ $? -ne 0 ]; then
     echo "cache file is not present in persistent path"
     exit 1
--- a/ci/sanity/sanity.sh
+++ b/ci/sanity/sanity.sh
@@ -127,6 +127,21 @@ kubectl apply -f ./ci/sanity/cspc.yaml
 
 IsPoolHealthy 30
 
+## Once the pool is healthy verify whether cachefile stored in persistent path
+## Get the CSPC name
+echo "Verifying whether pool cachefile is stored in persistent path"
+cspcName=$(kubectl get cspc -n openebs -o jsonpath={.items[0].metadata.name})
+if [ $? -ne 0 ]; then
+    echo "Failed to get CSPC name"
+    exit 1
+fi
+## verify whether cache file present in persistent path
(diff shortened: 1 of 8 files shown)


mittachaitu:
go.mod needs to be updated once PR got merged https://github.com/openebs/api/pull/48

ci/sanity/sanity.sh
```diff
@@ -127,10 +127,25 @@ kubectl apply -f ./ci/sanity/cspc.yaml
...
+    echo "Failed to get CSPC name"
+    exit 1
+fi
+## verify whether cache file present in persistent path
+ls -lrth /var/openebs/cstor-pool/${cspc_name}/pool.cache
+if [ $? -ne 0 ]; then
+    echo "cache file is not present in persistent path"
+    exit 1
+fi
+
```
sonasingh46, line 144: Awesome job for adding this to ci. :heart:

pkg/pool/operations/import.go
```diff
@@ -55,8 +55,10 @@ func (oc *OperationsConfig) Import(cspi *cstor.CStorPoolInstance) (bool, error)
 	devID := pool.GetDevPathIfNotSlashDev(bdPath[0])
 	cacheFile := types.CStorPoolBasePath + types.CacheFileName
 	if len(devID) != 0 {
+		// Cachefile property takes more precedence over the directory
+		// for sparse based pools importing based on directory make more
+		// sense. As if we mention WithCacheFile then directory is not considered
 		cmdOut, err = zfs.NewPoolImport().
-			WithCachefile(cacheFile).
```
mittachaitu, line 59: @kmova @sonasingh46 Removed `-c cachefile` option in the import command for sparse based pool. While building the command directory option will be ignored if cachefile is set.

sonasingh46: As discussed -- let us make the workflow :
1. Cache file import
2. Then, directory-based import if above fails

For all the pool type -- whether be sparse or disk.

mittachaitu: pushed the changes!! Verifying the same

shubham14bajpai: Changes look good.

PR #79, opened 2020-05-20
refact(cspc): add migration support for spc to cspc

shubham14bajpai:

This PR adds the code that supports migration from spc to cspc. The key changes made are:
 - while creating the cspc during migration an annotation `reconcile.openebs.io/disable-dependants` will set. This will help set reconciliation off on all of new cspi created.
 - modify the import command to rename the pool when importing the old pool.
   The command will be modified such that if the old pool name is not provided the regular import will work as expected. If old name is provided the command will look like this
```
/usr/local/bin/zpool import  -o cachefile=/var/openebs/cstor-poolpool.cache  -d /var/openebs/sparse cstor-08bfced5-6a28-4a63-a76b-c48c69be6ad5 cstor-6b3bbf9e-451d-4333-b119-1bb5217e5bc2
``` 
 otherwise it will be
```
/usr/local/bin/zpool import  -o cachefile=/var/openebs/cstor-poolpool.cache  -d /var/openebs/sparse cstor-6b3bbf9e-451d-4333-b119-1bb5217e5bc2
```

For more info refer: https://github.com/openebs/openebs/pull/2840


pkg/cspc/algorithm/build_csp.go
```diff
@@ -78,6 +78,13 @@ func (ac *Config) GetCSPISpec() (*cstor.CStorPoolInstance, error) {
 		WithFinalizer(types.CSPCFinalizer).
 		WithNewVersion(version.GetVersion()).
 		WithDependentsUpgraded()
+	// check for OpenEBSDisableDependantsReconcileKey annotation which implies
+	// the CSPI should have OpenEBSDisableReconcileLabelKey enabled
+	if ac.CSPC.GetAnnotations()[types.OpenEBSDisableDependantsReconcileKey] == "true" {
+		cspiObj.WithAnnotationsNew(map[string]string{
```
mittachaitu, line 84: Instead of `WithAnnotationsNew` use `WithAnnotations`. So in the future, if we add `WithannotationNew()` in the above builder we no need to change this line.

pkg/pool/operations/import.go
```diff
@@ -53,14 +56,18 @@ func (oc *OperationsConfig) Import(cspi *cstor.CStorPoolInstance) (bool, error)
 	klog.Infof("Importing pool %s %s", string(cspi.GetUID()), PoolName())
```
mittachaitu, line 56: Should we need to have a log saying that renaming the pool? It will help us to understand user is renaming existing pool.

pkg/pool/operations/import.go
```diff
@@ -53,14 +56,18 @@ func (oc *OperationsConfig) Import(cspi *cstor.CStorPoolInstance) (bool, error)
...
 	// Import the pool using cachefile
 	// command will looks like: zpool import -c <cachefile_path> -o <cachefile_path> <pool_name>
 	cmdOut, err = zfs.NewPoolImport().
 		WithCachefile(cacheFile).
 		WithProperty("cachefile", cacheFile).
-		WithPool(PoolName()).
-		WithExecutor(oc.zcmdExecutor).
+		WithPool(existingPoolName).
+		WithNewPool(PoolName()).
+    WithExecutor(oc.zcmdExecutor).
```
mittachaitu, line 70: indentation is missing

pkg/pool/operations/import.go
```diff
@@ -53,14 +56,18 @@ func (oc *OperationsConfig) Import(cspi *cstor.CStorPoolInstance) (bool, error)
...
 
 	// Import the pool using cachefile
 	// command will looks like: zpool import -c <cachefile_path> -o <cachefile_path> <pool_name>
 	cmdOut, err = zfs.NewPoolImport().
 		WithCachefile(cacheFile).
 		WithProperty("cachefile", cacheFile).
-		WithPool(PoolName()).
-		WithExecutor(oc.zcmdExecutor).
+		WithPool(existingPoolName).
+		WithNewPool(PoolName()).
```
mittachaitu, line 69: Good to put a comment of command if `existingPoolName` is set how it looks?

mittachaitu: Similarly in another place also

pkg/pool/operations/import.go
```diff
@@ -76,8 +83,9 @@ func (oc *OperationsConfig) Import(cspi *cstor.CStorPoolInstance) (bool, error)
 		cmdOut, err = zfs.NewPoolImport().
 			WithDirectory(devID).
 			WithProperty("cachefile", cacheFile).
-			WithPool(PoolName()).
-			WithExecutor(oc.zcmdExecutor).
+			WithPool(existingPoolName).
```
mittachaitu, line 86: Since it was scanning through all the blockdevices(If bad disk exists then migration will not be successful). Good to have a **note** saying that before performing migration make sure that there shouldn't be any bad disks attached to the node.
To make sure they can run `sudo fdisk -l` on all the CStor-pool pod nodes if command hung then there will is some problem.
cc: @sonasingh46

pkg/cspc/algorithm/build_csp.go
shubham14bajpai: Done

pkg/pool/operations/import.go
shubham14bajpai: Done

pkg/pool/operations/import.go
shubham14bajpai: Done

pkg/pool/operations/import.go
shubham14bajpai: Done

pkg/pool/operations/import.go
shubham14bajpai: Sure will add this to migration documentation.

shubham14bajpai:
> Things will go wrong in the below scenario:
> 
> * The assume pool is imported and successfully renamed the pool name on the disk but before the control plane removes **OpenEBSCStorExistingPoolName** annotation if pool pod is restarted then it will never import the pool.

If the pool pod restarts, the pool will already be imported with new name and the below code should take care of removing the annotation.
```go
func (oc *OperationsConfig) Import(cspi *cstor.CStorPoolInstance) (bool, error) {
	if poolExist := checkIfPoolPresent(PoolName(), oc.zcmdExecutor); poolExist {
		// If the pool is renamed and imported but the pool-mgmt restarts
		// for some reason then the annotation should be removed.
		delete(cspi.Annotations, string(types.OpenEBSCStorExistingPoolName))
		return true, nil
	}
```

mittachaitu:
> > Things will go wrong in the below scenario:
> > 
> > * The assume pool is imported and successfully renamed the pool name on the disk but before the control plane removes **OpenEBSCStorExistingPoolName** annotation if pool pod is restarted then it will never import the pool.
> 
> If the pool pod restarts, the pool will already be imported with new name and the below code should take care of removing the annotation.
> 
> ```go
> func (oc *OperationsConfig) Import(cspi *cstor.CStorPoolInstance) (bool, error) {
> 	if poolExist := checkIfPoolPresent(PoolName(), oc.zcmdExecutor); poolExist {
> 		// If the pool is renamed and imported but the pool-mgmt restarts
> 		// for some reason then the annotation should be removed.
> 		delete(cspi.Annotations, string(types.OpenEBSCStorExistingPoolName))
> 		return true, nil
> 	}
> ```

This will help only if `cspi-mgmt` restarts. FYI for every restart of the `cstor-pool` container pool-mgmt must execute `zpool import` command

shubham14bajpai:
> > > Things will go wrong in the below scenario:
> > > 
> > > * The assume pool is imported and successfully renamed the pool name on the disk but before the control plane removes **OpenEBSCStorExistingPoolName** annotation if pool pod is restarted then it will never import the pool.
> > 
> > 
> > If the pool pod restarts, the pool will already be imported with new name and the below code should take care of removing the annotation.
> > ```go
> > func (oc *OperationsConfig) Import(cspi *cstor.CStorPoolInstance) (bool, error) {
> > 	if poolExist := checkIfPoolPresent(PoolName(), oc.zcmdExecutor); poolExist {
> > 		// If the pool is renamed and imported but the pool-mgmt restarts
> > 		// for some reason then the annotation should be removed.
> > 		delete(cspi.Annotations, string(types.OpenEBSCStorExistingPoolName))
> > 		return true, nil
> > 	}
> > ```
> 
> This will help only if `cspi-mgmt` restarts. FYI for every restart of the `cstor-pool` container pool-mgmt must execute `zpool import` command.
> 
> Below snippet will help to solve the problem
> https://github.com/openebs/maya/blob/e5f6235f716bbefe719f6d346b9125941bc1ca01/cmd/cstor-pool-mgmt/pool/v1alpha2/import.go#L50

Nice Catch :+1: Addressed the issue.

pkg/pool/operations/import.go
```diff
@@ -36,14 +36,30 @@ import (
...
+		delete(cspi.Annotations, types.OpenEBSCStorExistingPoolName)
 		return true, nil
 	}
 
+	var poolImported, importable bool
+	// existingPoolName denotes the pool name that may be present
+	// from previous version and needs to be imported with new name
+	existingPoolName := cspi.Annotations[types.OpenEBSCStorExistingPoolName]
+	if existingPoolName != "" {
+		_, importable, _ = oc.checkIfPoolIsImportable(cspi)
```
prateekpandey14, line 50: do we need to handle the return err here ?

shubham14bajpai: No if err occurs it will fail to import and re-attempt.

mittachaitu: Good to handle but it might not be mandatory.

prateekpandey14: that's correct , but its better to fail fast

PR #82, opened 2020-05-22
fix(webhook): fix blockdevice replacement validation

mittachaitu:
This PR fixes the blockdevice replacement validations performed by the CStor admission server.

**Old RaidGroups**: RaidGroups of CSPC present in etcd.
**New RaidGroups**: RaidGroups of newly requested CSPC for updates.

**Before bug fix**:
- We are forming an index mapping(two different RaidGroup lists let us say rgList1 and rgList2. In these lists rgList1 of 0th index and rgList2 of 0th index contains the same RaidGroup[Marked RaidGroup as common which has at least one common blockdevice]) between old and new RaidGroups.
- Iterating over rgList1 and rgList2(as an outer and inner loop) and checking whether it is a case of replacement. If it a case of replacement then further [validations](https://github.com/openebs/openebs/blob/master/contribute/design/1.x/cstor-operator/doc.md#disk-replacement) were made on the request.
- So when iterating over the second index on rgList1(outer loop) in the inner loop again starts from 0th index which leads to wrong validations.

**How This PR helps in bug fix**:
- Instead of iterating over two rgList1 and rgList2 iterate over only one list since both are already mapped with an index so it will have all common RaidGroups(old & new). For each iteration corresponding index will be picked and validated.

Error facing in replacing blockdevices other that first RaidGroup
```sh
error: cstorpoolclusters.cstor.openebs.io "cspc-stripe" could not be patched: admission webhook "admission-webhook.cstor.openebs.io" denied the request: invalid cspc specification: cannot replace more than one blockdevice in a raid group
You can run `kubectl replace -f /tmp/kubectl-edit-isowr.yaml` to try this update again.
```

### DIFF
--- a/pkg/webhook/cspc_operations.go
+++ b/pkg/webhook/cspc_operations.go
@@ -651,23 +651,21 @@ func (pOps *PoolOperations) ArePoolSpecChangesValid(oldPoolSpec, newPoolSpec *cs
 		rgType := v.rgType
 		for index, _ := range v.oldRaidGroups {
 			oldRg := v.oldRaidGroups[index]
-			for _, _ = range v.newRaidGroups {
-				newRg := v.newRaidGroups[index]
-				if err = validateRaidGroupChanges(&oldRg, &newRg, rgType); err != nil {
-					return false, fmt.Sprintf("raid group validation failed: %v", err)
+			// Already mapped(via index) old raid groups and new raid groups in
+			// commonRaidGroups no need to iterate over v.newRaidGroups
+			newRg := v.newRaidGroups[index]
+
+			if err = validateRaidGroupChanges(&oldRg, &newRg, rgType); err != nil {
+				return false, fmt.Sprintf("raid group validation failed: %v", err)
+			}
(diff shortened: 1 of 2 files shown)


pkg/webhook/cspc_operations.go
```diff
@@ -649,10 +649,10 @@ func (pOps *PoolOperations) ArePoolSpecChangesValid(oldPoolSpec, newPoolSpec *cs
 	}
 	for _, v := range commonRaidGroups {
 		rgType := v.rgType
-		for _, oldRg := range v.oldRaidGroups {
-			oldRg := oldRg
-			for _, newRg := range v.newRaidGroups {
-				newRg := newRg
+		for index, _ := range v.oldRaidGroups {
+			oldRg := v.oldRaidGroups[index]
+			for _, _ = range v.newRaidGroups {
```
sonasingh46, line 654: Do we need the for loop here?

mittachaitu: No need..

PR #89, opened 2020-06-03
refact(cvc-operator): update cvc controller for migration

shubham14bajpai:

This PR 
 - comments out the check for `nodeID` to allow newly created CVC to reconcile for a migrated volume.
 - adds code to honour the replica distribution for pool names mentioned in the CVC spec.

### DIFF
--- a/pkg/controllers/cstorvolumeconfig/controller.go
+++ b/pkg/controllers/cstorvolumeconfig/controller.go
@@ -208,18 +208,18 @@ func (c *CVCController) syncCVC(cvc *apis.CStorVolumeConfig) error {
 		return nil
 	}
 
-	nodeID := cvc.Publish.NodeID
-	if nodeID == "" {
-		// We choose to absorb the error here as the worker would requeue the
-		// resource otherwise. Instead, the next time the resource is updated
-		// the resource will be queued again.
-		runtime.HandleError(fmt.Errorf("cvc must be publish/attached to Node: %+v", cvc))
-		c.recorder.Event(cvc, corev1.EventTypeWarning,
-			Provisioning,
-			fmt.Sprintf(MessageCVCPublished, cvc.Name),
-		)
-		return nil
-	}
+	// nodeID := cvc.Publish.NodeID
+	// if nodeID == "" {
+	// 	// We choose to absorb the error here as the worker would requeue the
+	// 	// resource otherwise. Instead, the next time the resource is updated
+	// 	// the resource will be queued again.
+	// 	runtime.HandleError(fmt.Errorf("cvc must be publish/attached to Node: %+v", cvc))
+	// 	c.recorder.Event(cvc, corev1.EventTypeWarning,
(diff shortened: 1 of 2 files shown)


pkg/controllers/cstorvolumeconfig/volume_operations.go
```diff
@@ -341,6 +359,10 @@ func (c *CVCController) distributeCVRs(
 		return err
 	}
 
+	if len(policy.Spec.ReplicaPoolInfo) != 0 {
```
mittachaitu, line 362: Should we need to add strict check?
```go
if len(policy.Spec.ReplicaPoolInfo) != 0
if len(policy.Spec.ReplicaPoolInfo) != claim.Spec.Provision.ReplicaCount {
// return error
}
getPolicyBasedPoolList()
}
```

pkg/controllers/cstorvolumeconfig/controller.go
```diff
@@ -208,18 +208,18 @@ func (c *CVCController) syncCVC(cvc *apis.CStorVolumeConfig) error {
...
-		// resource otherwise. Instead, the next time the resource is updated
-		// the resource will be queued again.
-		runtime.HandleError(fmt.Errorf("cvc must be publish/attached to Node: %+v", cvc))
-		c.recorder.Event(cvc, corev1.EventTypeWarning,
-			Provisioning,
-			fmt.Sprintf(MessageCVCPublished, cvc.Name),
-		)
-		return nil
-	}
+	// nodeID := cvc.Publish.NodeID
```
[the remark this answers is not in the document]
prateekpandey14: Yes its fine to remove.

pkg/controllers/cstorvolumeconfig/volume_operations.go
shubham14bajpai: Done

pkg/controllers/cstorvolumeconfig/volume_operations.go
```diff
@@ -312,6 +312,24 @@ func (c *CVCController) getOrCreateCStorVolumeResource(
...
+	poolInfo []apis.ReplicaPoolInfo,
+) *apis.CStorPoolInstanceList {
+	policyBasedList := &apis.CStorPoolInstanceList{}
+	for _, info := range poolInfo {
+		for _, pool := range poolList.Items {
+			if pool.Name == info.PoolName {
+				policyBasedList.Items = append(
+					policyBasedList.Items,
+					pool,
+				)
```
mittachaitu, line 326: add `break` over here.

shubham14bajpai: If we add break here won't it return only one pool name in the list? We want all the pool names in the policy.

shubham14bajpai: Ok got it. Ignore my comment.

PR #93, opened 2020-06-11
chore(cvc): add cvc-controller fake client mock tests

prateekpandey14:

### DIFF
--- a/pkg/controllers/cstorvolumeconfig/controller_test.go
+++ b/pkg/controllers/cstorvolumeconfig/controller_test.go
@@ -17,15 +17,340 @@ limitations under the License.
 package cstorvolumeconfig
 
 import (
+	"fmt"
+	"html"
 	"reflect"
+	"strconv"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/controllers/cstorvolumeconfig/controller_test.go
```diff
@@ -17,16 +17,341 @@ limitations under the License.
...
+	}
+	for _, cvc := range f.cvcLister {
+		cvcInformerFactory.Cstor().V1().CStorVolumeConfigs().Informer().GetIndexer().Add(cvc)
+	}
+
+	for _, cvr := range f.cvrLister {
+		cvcInformerFactory.Cstor().V1().CStorVolumeReplicas().Informer().GetIndexer().Add(cvr)
+	}
+
+	return c, cvcInformerFactory, recorder, nil
```
mittachaitu, line 214: nit: Instead of returning `recorder` we can typecast recorder interface in controller struct `c.recorder.(*record.FakeRecorder)`

prateekpandey14: yes, i will change that in upcoming PRs with some new tests

PR #99, opened 2020-06-12
feat(cspi): add provisioned and healthy replica count in CSPI status

mittachaitu:
This PR adds Provisioned and Healthy
replica count present in the pool to CSPI status.
This will be helpful to the users to know no.of
replicas are present in the pool and out of them
`Healthy` replica count.

- When users execute `kubectl get cspi -n openebs` below will be output
```sh
 kubectl get cspi -n openebs
NAME                HOSTNAME    ALLOCATED   FREE     CAPACITY    READONLY   PROVISIONEDREPLICAS   HEALTHYREPLICAS   STATUS   AGE
cstor-stripe-fb4b   127.0.0.1   190k        19200M   19200190k   false      2                     1                 ONLINE   36m
```
**Provisioned Replicas**:  Count of volume replicas present in the pool.
**Healthy Replicas**: Count of Healthy volume replicas present in the pool.

Users will able  to know provisioned and healthy replica count 
without executing extra command `kubectl get cvr -n openebs -l cstorpoolinstance.openebs.io/name=<cspi_name>`

**Note to reviewers**:
- `Provisioned Replicas` is calculated from the output of
   `zfs list` after filtering clones and pool name.
- `Healthy Replicas` is calculated from using `zfs stats`
   which is json output.
- `ZFS Stats` will give the status information of replica
   only if the dataset has target IP. If we get provisioned
   count of replicas from stats then there will be incorrect
   output when user is restoring the volumes to mitigate
   this we are fetching output from `zfs list`(which will give
   all the datasets). 

Corresponding API repo PR: https://github.com/openebs/api/pull/54


mittachaitu:
Sanity test needs to be enable in Travis

mittachaitu:
> Sanity test needs to be enable in Travis

Pushed commit to uncomment sanity test which was commented in PR #104

PR #101, opened 2020-06-17
feat(webhook): add cspc scaledown validations

shubham14bajpai:

This PR adds cspc scale down validations. The webhook validates scale down by looking for any CVR exists on the scaled-down pool.

### DIFF
--- a/pkg/webhook/cspc.go
+++ b/pkg/webhook/cspc.go
@@ -492,6 +492,9 @@ func (wh *webhook) validateCSPCUpdateRequest(req *v1beta1.AdmissionRequest, getC
 
 	if ok, msg := pOps.ValidateScaledown(); !ok {
 		err = errors.Errorf("invalid cspc specification: %s", msg)
+		// As scale down validation may take more time than the timeout value set
+		// on webhook having a log will help debug
+		klog.Error(err)
 		response = BuildForAPIObject(response).UnSetAllowed().WithResultAsFailure(err, http.StatusUnprocessableEntity).AR
 		return response
 	}
--- a/pkg/webhook/cspc.go
+++ b/pkg/webhook/cspc.go
@@ -32,6 +32,7 @@ import (
 	"github.com/pkg/errors"
 	"k8s.io/api/admission/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/klog"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/webhook/cspc.go
```diff
@@ -488,6 +489,13 @@ func (wh *webhook) validateCSPCUpdateRequest(req *v1beta1.AdmissionRequest, getC
 		return response
 	}
 	pOps := NewPoolOperations(wh.kubeClient, wh.clientset).WithNewCSPC(&cspcNew).WithOldCSPC(cspcOld)
+
+	if ok, msg := pOps.ValidateScaledown(); !ok {
+		err = errors.Errorf("invalid cspc specification: %s", msg)
```
mittachaitu, line 494: Good to log after this line so that it will be helpful for debugging

pkg/webhook/cspc.go
```diff
@@ -503,3 +511,63 @@ func (wh *webhook) validateCSPCUpdateRequest(req *v1beta1.AdmissionRequest, getC
...
+				})
+			if err != nil {
+				return false, fmt.Sprintf("Could not list cspi for cspc %s: %s", p.OldCSPC.Name, err.Error())
+			}
+			removedPools = append(removedPools, cspi.Items[0].Name)
+		}
+	}
+	for _, cspiName := range removedPools {
+		// list cvrs in all namespaces
+		cvrList, err := p.clientset.CstorV1().CStorVolumeReplicas("").List(metav1.ListOptions{
```
prateekpandey14, line 541: no need to list in all-namespace here as pool is already exists, we can use the cspc `Namespace`.

mittachaitu: Since it is a single n/w call I believe there won't be any problem with that if the thought was about payload then we can do that.

shubham14bajpai: Done

pkg/webhook/cspc.go
shubham14bajpai: As this error will be displayed on the console when scaledown validation fails and the patch or update done by the user will not be executed so no need for log here like other cases.

mittachaitu: It will help in debugging time.

shubham14bajpai: As the operation was not successful (the current state of cspc did not change) and the user can see the error in front of him how will logging the same error help?

shubham14bajpai: Done

PR #105, opened 2020-06-24
cherry-pick(VERSION): cherry-pick of #104 and #106

mittachaitu:
- This PR is cherry-pick of https://github.com/openebs/cstor-operators/pull/104 and https://github.com/openebs/cstor-operators/pull/106
which fixes the version in labels and also uncomments the tests in Travis(which
was commented in #104).
- For more info about the PR have look at the description of PR https://github.com/openebs/cstor-operators/pull/104 and https://github.com/openebs/cstor-operators/pull/106


build/build.sh
```diff
@@ -39,7 +49,25 @@ if [[ -n "$TRAVIS_TAG" ]] && [[ $TRAVIS_TAG != *"RC"* ]]; then
...
+## If travis tag is set then assign it as VERSION and
+## if travis tag is empty then mark version as ci
+if [ -n "$TRAVIS_TAG" ]; then
+    # Trim the `v` from the TRAVIS_TAG if it exists
+    # Example: v1.10.0 maps to 1.10.0
+    # Example: 1.10.0 maps to 1.10.0
+    # Example: v1.10.0-custom maps to 1.10.0-custom
+    VERSION="${TRAVIS_TAG#v}"
+else
+    VERSION="dev"
```
pawanpraka1, line 64: let's make it VERSION="$CURRENT_BRANCH-$SHORT_COMMIT:$BUILDDATE".

build/build.sh
```diff
@@ -39,7 +49,25 @@ if [[ -n "$TRAVIS_TAG" ]] && [[ $TRAVIS_TAG != *"RC"* ]]; then
...
+fi
+
+echo "Building for VERSION ${VERSION}"
+## Below line will help to get current version for various binaries
+echo "${VERSION}" > "$VERSION_FILE_PATH"
```
pawanpraka1, line 69: why do we need version file now? Just setting the VERSION variable should be sufficient.

build/build.sh
mittachaitu: - This VERSION is used to represent the version of OpenEBS via labels and
version details(IMO going with commit based doesn't fit here).
- Fetching [version](https://github.com/openebs/cstor-operators/blob/3b13054649044aae928c258ee7c160dea8254bb6/pkg/version/version.go#L107) from file and we are fetching version using `GetVersion` in multiple places.

build/build.sh
mittachaitu: In code, we are making use of the VERSION file because of this reason generated only during build time and deleted after that.

pawanpraka1: can you show me where we are using?

build/build.sh
pawanpraka1: GetVersion is executed runtime , this file should be present in the docker file. We don't need this file.

pawanpraka1: >This VERSION is used to represent the version of OpenEBS via labels and
version details(IMO going with commit based doesn't fit here).
for development, we can have the branch and the commit hash from which we have comiled it code. After this change all the local compiled code will show dev as VERSION.

build/build.sh
mittachaitu: image

build/build.sh
mittachaitu: > GetVersion is executed runtime , this file should be present in the docker file. We don't need this file.

We are statically linking during compilation time https://github.com/openebs/cstor-operators/blob/3b13054649044aae928c258ee7c160dea8254bb6/build/build.sh#L133

pawanpraka1: sai that is variable, file was just used to set VERSION variable. The file is not used anywhere else.

mittachaitu: Correct I also just now realized while discussing with Prateek.

mittachaitu: >After this change all the local compiled code will show dev as VERSION.

Correct!! So how about having branch_name-dev i.e if the branch is v1.11.x then label will be `1.11.0-dev`

build/build.sh
mittachaitu: > why do we need version file now? Just setting the VERSION variable should be sufficient.

Correct!! setting the VERSION variable is good enough!!

build/build.sh
mittachaitu: The reason why I am not going via commit based is we have few [checks](https://github.com/openebs/cstor-operators/blob/3b13054649044aae928c258ee7c160dea8254bb6/pkg/controllers/cstorvolumeconfig/controller.go#L847) in the code if checks mismatch then reconciliation will not happen.

`CVC.VersionDetails.Status.Current` will be populated by [CSI-Plugin](https://github.com/openebs/cstor-csi/blob/68bb336e9777f139127b75c79b5be55dfee906dd/pkg/utils/maya.go#L110) if it mismatch then CStorVolume resources will not be created.

pawanpraka1: did not get that check part entirly, but lets us go ahead with branch_name-dev.

mittachaitu: Basically that check was added to stop dataplane upgrade when control plane alone is upgraded. I will go ahead with branch_name-dev

mittachaitu: Raised a PR https://github.com/openebs/cstor-operators/pull/106 after PR got merge then I will cherry-pick commit into this PR

build/build.sh
```diff
@@ -17,6 +17,16 @@
 # This script builds the application from source for multiple platforms.
 set -e
 
+VERSION_FILE_PATH="$GOPATH/src/github.com/openebs/cstor-operators/VERSION"
```
kmova, line 20: Based on the below comments, it will be good to delete this file. Also helps with cases where builds need to happen on forked repos.

mittachaitu: Yes, I have a PR #106 to fix

build/build.sh
```diff
@@ -39,7 +39,36 @@ if [[ -n "$TRAVIS_TAG" ]] && [[ $TRAVIS_TAG != *"RC"* ]]; then
 fi
 
 # Get the version details
-VERSION="$(cat $GOPATH/src/github.com/openebs/cstor-operators/VERSION)"
+#VERSION="$(cat $GOPATH/src/github.com/openebs/cstor-operators/VERSION)"
```
prateekpandey14, line 42: remove this comment itself

PR #106, opened 2020-06-25
fix(VERSION): omit generation of VERSION file

mittachaitu:
This PR does the following changes:
- Omits the VERSION file generation.
- Populates the current and desired version details for CVC if it is empty.

VERSION can have the following possible values:
- If Travis_TAG is set then `TRAVIS_TAG` then corresponding
   tag will be the value.
- If TRAVIS_TAG is not set then the value will be  `${branch_name}-dev`

**Why we need this PR**:
This PR will populate the value of `openebs.io/version` and internal
resource version details based on build tag and build branch.

**Special Note To Reviewers**:
- VERSION value will be set on internal resources created by CVC and CSPC operators.
- This PR is continuation of #104 

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -60,20 +60,11 @@ if [ -n "$TRAVIS_TAG" ]; then
     # Example: v1.10.0-custom maps to 1.10.0-custom
     VERSION="${TRAVIS_TAG#v}"
 else
-    ## Counting dots in CURRENT_BRANCH. If it is
-    ## release branch then it will have more than
-    ## one dot(.) because we are following release
-    ## branch naming convention as v1.10.x, v0.9.x
-    ## v2.0.x, v1.0.x
-
-    ## If we compare with master then local building
-    ## will have unknown version
-    dot_count=$(echo "${CURRENT_BRANCH}" | grep -o "\." | wc -l)
-    if [ "${dot_count}" -gt 1 ]; then
-        VERSION="${CURRENT_BRANCH#v}-dev"
-    else
-        VERSION="dev"
-    fi
+    ## Marking VERSION as current_branch-dev
+    ## Example: master branch maps to master-dev
+    ## Example: v1.11.x-ee branch to 1.11.x-ee-dev
+    ## Example: v1.10.x branch to 1.10.x-dev
+    VERSION="${CURRENT_BRANCH#v}-dev"
 fi
 
 echo "Building for ${VERSION} VERSION"
--- a/build/build.sh
+++ b/build/build.sh
@@ -17,16 +17,6 @@
 # This script builds the application from source for multiple platforms.
 set -e
 
-VERSION_FILE_PATH="$GOPATH/src/github.com/openebs/cstor-operators/VERSION"
-
-on_exit() {
-    ## Delete VERSION file
-    echo "Deleteing VERSION File($VERSION_FILE_PATH) that got generated during build time"
-    rm -f "$VERSION_FILE_PATH"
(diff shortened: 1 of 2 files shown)


build/build.sh
```diff
@@ -61,12 +60,23 @@ if [ -n "$TRAVIS_TAG" ]; then
...
-    VERSION="dev"
+    ## Counting dots in CURRENT_BRANCH. If it is
+    ## release branch then it will have more than
+    ## one dot(.) because we are following release
+    ## branch naming convention as v1.10.x, v0.9.x
+    ## v2.0.x, v1.0.x
+
+    ## If we compare with master then local building
+    ## will have unknown version
+    dot_count=$(echo "${CURRENT_BRANCH}" | grep -o "\." | wc -l)
```
kmova, line 71: Can you post what the version will look like for the following branches:
- v1.11.x
- v1.11.x-ee
- v1.11.0
- v1.11.0-RC1

mittachaitu: If we raise a PR against below branch names then corresponding VERSION values
**BRANCH_NAME**    ---     **VERSION**
```sh
v1.11.x                          ---     1.11.x-dev
v1.11.x-ee                     ---     1.11.x-ee-dev
v1.11.0                          ---     1.11.0-dev
v1.11.0-RC1                 ---     1.11.0-RC1-dev
```
Above **VERSION** will be generated only if TRAVIS_TAG is [empty](https://github.com/openebs/cstor-operators/blob/3b13054649044aae928c258ee7c160dea8254bb6/build/build.sh#L57).

`v1.11.0 and v1.11.0-RC1` are tags. So when TRAVIS_TAG is set then VERSION value will be described as below
**TRAVIS_TAG**    ---     **VERSION**
```sh
v1.11.0-RC1            ---     1.11.0-RC1
v1.11.0                    ---      1.11.0    
```

build/build.sh
```diff
@@ -61,12 +60,23 @@ if [ -n "$TRAVIS_TAG" ]; then
...
+    if [ "${dot_count}" -gt 1 ]; then
```
pawanpraka1, line 72: what is the need to do this? why not just do VERSION="${CURRENT_BRANCH#v}-ci" non tagged build?

mittachaitu: Basically following is the flow when the user creates  PVC:
- CSI driver receives volume create requests and creates CVC(With all required details and it will populate version also -- We need to fix it but I will take in the current release but this PR will go to `kubera`).
- CVC controller will reconcile for CVCs only if `cvc.VersionDetails.Status.Current` matches to VERSION value in CVC controller.

If we follow VERSION value as `${CURRENT_BRANCH#v}-ci` then if any contributor builds the code and tries to test volume provisioning then it will fail due to version mismatch.

pawanpraka1: here in this case also the same problem is there. Say CSI version he is using is 1.11.0 and when he will comile the code locally if
1. the branch is v1.11.x,  version will be 1.11.x-dev
2. the branch is master, then version will be dev 

He has to compile all the dependant repo to test his change?

mittachaitu: > He has to compile all the dependant repo to test his change?

Yes(Need to remove dependency). I raised an issue https://github.com/openebs/cstor-operators/issues/108 which will be targeted for this release.

mittachaitu: > the branch is v1.11.x, version will be 1.11.x-dev
I was in a thought that contributor will checkout from release branch and create tags then build it.

PR #110, opened 2020-06-29
refact(webhook): make webhook config failure policy configurable

prateekpandey14:
**What this PR does?:**

commit enable the webhook validatingwebhookconfiguration failure policy configurable using a env called
ADMISSION_WEBHOOK_FAILURE_POLICY.

There are 2 types of failure policy which can be configurable are `Fail` and `Ignore`. `Fail` will be the default policy

**Does this PR require any upgrade changes?**:
NO

**If the changes in this PR are manually verified, list down the scenarios covered:**:
- deployed the admission server using `ignore` failure policy and verified that it got configured properly

### DIFF
--- a/pkg/webhook/configuration.go
+++ b/pkg/webhook/configuration.go
@@ -34,6 +34,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/webhook/configuration.go
```diff
@@ -210,7 +214,8 @@ func (c *client) createAdmissionValidatingConfig(
 		// SideEffects:             &sideEffectClass,
 		// AdmissionReviewVersions: []string{"v1"},
 		TimeoutSeconds: &five,
-		FailurePolicy:  &Fail,
+		//FailurePolicy:  &Fail,
+		FailurePolicy: failurePolicy(),
```
mittachaitu, line 218: If the user set policy as `Ignore` in env and upgraded to 1.12.0 thinking they marked `FailurePolicy` as ignore but it is not the case.

prateekpandey14: No, they have to delete the valdatingwebhookconfiguration and set the ENV which will restart the admission pod and config will be created again.

sonasingh46: As part of upgrade -- can't we take this part also into account and upgrade job should/can handle this. @mittachaitu @prateekpandey14

kmova: n_ : remove the commented line. 

Also, as this was in alpha, we are ok to make this change.

prateekpandey14: commented line removed

PR #114, opened 2020-07-01
test(restore): add mock test for restore endpoint

mittachaitu:
This PR adds a mock test(unit test) for `Restore` endpoint.

_List of unit test_
**Backup endpoint**
- Trigger delete method on the backup endpoint(It should delete snapshot, CStorBackup, and CStorCompletedBackup CR's).
- Trigger delete method on the backup endpoint by injecting errors in snapshot deletion(should return an error).
- Trigger delete method on the backup endpoint when the backup is for scheduled backup(It should not delete CStorCompletedBackup).
- Trigger delete method on the backup endpoint without the required information(should return an error).

**Restore endpoint**
- Trigger POST method for remote restore on the restore endpoint(It should create required resources for restore).
- Trigger POST method for remote restore on the restore endpoint by disabling the CVC reconciliation(should return an error).
- Trigger POST method for local restore on the restore endpoint(positive and error injection are covered).

### DIFF
--- a/pkg/server/cstorvolumeconfig/https_test.go
+++ b/pkg/server/cstorvolumeconfig/https_test.go
@@ -190,7 +190,7 @@ func (f *fixture) createFakeVolumeReplicas(
 			WithLabelsNew(labels).
 			WithStatusPhase(phase)
 		_, err := f.openebsClient.CstorV1().CStorVolumeReplicas(namespace).Create(cvr)
-		if err != nil && !k8serror.IsNotFound(err) {
+		if err != nil && !k8serror.IsAlreadyExists(err) {
 			return err
 		}
 	}
@@ -207,7 +207,7 @@ func (f *fixture) createFakeCStorVolume(volumeName string) error {
 		WithName(volumeName).
 		WithLabelsNew(labels)
 	_, err := f.openebsClient.CstorV1().CStorVolumes(namespace).Create(cv)
-	if err != nil && !k8serror.IsNotFound(err) {
+	if err != nil && !k8serror.IsAlreadyExists(err) {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/server/cstorvolumeconfig/https_test.go
```diff
@@ -182,11 +186,11 @@ func (f *fixture) createFakeVolumeReplicas(
 			"cstorvolume.openebs.io/name":              volumeName,
 		}
 		cvr := cstor.NewCStorVolumeReplica().
-			WithName(volumeName + "-" + cspiList.Items[0].Name).
+			WithName(volumeName + "-" + cspiList.Items[i].Name).
 			WithLabelsNew(labels).
 			WithStatusPhase(phase)
 		_, err := f.openebsClient.CstorV1().CStorVolumeReplicas(namespace).Create(cvr)
-		if err != nil {
+		if err != nil && !k8serror.IsNotFound(err) {
```
prateekpandey14, line 193: why there will be not found error while create ?

pkg/server/cstorvolumeconfig/https_test.go
```diff
@@ -199,12 +203,14 @@ func (f *fixture) createFakeCStorVolume(volumeName string) error {
 		openebstypes.PersistentVolumeLabelKey: volumeName,
 	}
 	cv := cstor.NewCStorVolume().
+		WithNamespace(namespace).
 		WithName(volumeName).
 		WithLabelsNew(labels)
 	_, err := f.openebsClient.CstorV1().CStorVolumes(namespace).Create(cv)
-	if err != nil {
+	if err != nil && !k8serror.IsNotFound(err) {
```
prateekpandey14, line 210: why there will be not found error while create ?

pkg/server/cstorvolumeconfig/https_test.go
mittachaitu: It shouldn't be`IsAlreadyExist` somehow copy-pasted the wrong check

pkg/server/cstorvolumeconfig/https_test.go
mittachaitu: same as above comment btw updated the check

PR #122, opened 2020-07-09
fix(cstor-webhook): fixes pool scaledown bug

sonasingh46:
In case of scaling down a pool by removing a pool spec from the CSPC, the webhook server was reporting an out of bounds error. This commit fixes this bug.

**Bug Description:**  
The validation code, in case of scaledown, was trying to list CSPI based on all the labels present on the nodes of the corresponding CSPI. On the other hand, CSPI only has `kubernetes.io/hostname` label to it and not all the other node labels. As labels are ANDed for output, we were not getting any items in the CSPI list and hence out of bounds error was caused.

### DIFF
--- a/pkg/webhook/cspc.go
+++ b/pkg/webhook/cspc.go
@@ -28,7 +28,7 @@ import (
 	openebsapis "github.com/openebs/api/pkg/apis/openebs.io/v1alpha1"
 	"github.com/openebs/api/pkg/apis/types"
 	clientset "github.com/openebs/api/pkg/client/clientset/versioned"
-	util "github.com/openebs/api/pkg/util"
+	"github.com/openebs/api/pkg/util"
 	"github.com/pkg/errors"
 	"k8s.io/api/admission/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -519,6 +519,7 @@ func (wh *webhook) validateCSPCUpdateRequest(req *v1beta1.AdmissionRequest, getC
 // that is being scaled down
 func (p *PoolOperations) ValidateScaledown() (bool, string) {
 	removedPools := []string{}
+
 	for _, oldPool := range p.OldCSPC.Spec.Pools {
 		found := false
 		for _, newPool := range p.NewCSPC.Spec.Pools {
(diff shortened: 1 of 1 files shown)


pkg/webhook/cspc.go
```diff
@@ -527,11 +528,18 @@ func (p *PoolOperations) ValidateScaledown() (bool, string) {
 				break
 			}
 		}
+
+		ls := &metav1.LabelSelector{
```
mittachaitu, line 532: Can we move the initialization part inside the `if` block? (It will be more readable since we require this only if it is scale down case)

pkg/webhook/cspc.go
```diff
@@ -527,11 +528,18 @@ func (p *PoolOperations) ValidateScaledown() (bool, string) {
...
+			MatchLabels: map[string]string{
+				types.CStorPoolClusterLabelKey: p.OldCSPC.Name,
+				types.HostNameLabelKey:         oldPool.NodeSelector[types.HostNameLabelKey],
```
mittachaitu, line 535: Since `NodeSelector` is populated by `[Users]`(https://github.com/openebs/cstor-operators/blob/74b6a3a04e58b45136bbb077ee1dcba0f3c47e1b/pkg/cspc/algorithm/build_csp.go#L71) it will not guarantee that we can fetch `kubernetes.io/hostName` value using NodeSelector(Thinking about a case where users provide some other unique key to identify the node).

pkg/webhook/cspc.go
sonasingh46: Done. Thanks

pkg/webhook/cspc.go
sonasingh46: The `kubernetes.io/hostname` label -- whether provided by the user or not -- finally CSPC-operator finds out the node from the labels provided in the CSPC and adds a `kubernetes.io/hostname` label on CSPI by itself.

sonasingh46: But, will relook on that workflow and test it. If there is any fix required -- that will go in cspc-operator.

sonasingh46: Got your point... This is a nice catch. Done

prateekpandey14: Good catch !!!

Issue #134, opened 2020-07-20
bug(scaledown): admission server rejects update request of CSPC


mittachaitu:
- Created cStor pool using CSPC by populating specs.
- One of the node the cluster was lost and tried to remove corresponding node spec from CSPC.
- Admission server rejects the request by throwing below error:
```sh
could not find common pool specs for validation: invalid no.of nodes 0 from the given node selectors
```
Above error is from [function](https://github.com/openebs/cstor-operators/blob/0dda3cf9525c6ba8313014a860b766b369506a41/pkg/webhook/cspc_operations.go#L122). 

Expected:
- Should be able to remove the pool spec from CSPC when node doesn't exists.

PR #135, opened 2020-07-20
fix(cstor-webhook): allow scale down of pools when node doesn't exist in cluster

mittachaitu:
**What does this PR does**:
In case of scaling down a pool by removing a pool spec
from the CSPC when a pool spec node doesn't exist in
the cluster, the cStor-admission server reported an error
saying no node exists for a given node selector. This PR
fixes this bug.

**Bug Description**:
- To validate day-2-operations(pool expansion, blockdevice replacement)
   performed on the pool it was trying to get common pool specs from old
   CSPC spec(CSPC exist in etcd) and new CSPC spec(modified CSPC
   which needs to persist in etcd). common pool specs were figured out
   by fetching hostName from node resource(node resource will be fetched
   from etcd using nodeSelector on pool specs). In node doesn't exist
   in the cluster then scaling down operation will be failed.

**Issue fixed by this PR** : https://github.com/openebs/cstor-operators/issues/134

### DIFF
--- a/pkg/webhook/cspc_test.go
+++ b/pkg/webhook/cspc_test.go
@@ -1228,7 +1228,7 @@ func TestCSPCScaleDown(t *testing.T) {
 			expectedRsp: true,
 			getCSPCObj:  getCSPCObject,
 		},
-		"Scale down when node doesn't exist in cluster": {
+		"Positive scale down pool when node doesn't exist in cluster": {
 			existingObj: cstor.NewCStorPoolCluster().
 				WithName("cspc-bar-mirror-2").
 				WithNamespace("openebs").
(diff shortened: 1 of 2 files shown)


mittachaitu:
`pr/hold-merge` has been added to discusses corner cases.

pkg/webhook/cspc_test.go
```diff
@@ -1128,6 +1228,72 @@ func TestCSPCScaleDown(t *testing.T) {
...
+							),
+						),
+				),
+			existingCSPIs: []*cstor.CStorPoolInstance{
+				{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "cspc-bar-mirror-2-cspi-1",
+						Namespace: "openebs",
+						Labels: map[string]string{
+							types.HostNameLabelKey:         "worker-not-exist",
```
sonasingh46, line 1278: How does this simulate -- node does not exist behavior?

mittachaitu: First, we persisted CSPC in etcd with a pool spec belongs to a node that doesn't exist in etcd. During CSPC update time we removed the pool spec belongs to the node that it doesn't exist in the cluster.

At very high-level this test scenario is like:
- CSPC provisioned the pools on node1, node2, and node3 it interns creates CSPIs on specified nodes.
- Now let us assume one of the nodes lets say node2 has been removed from the cluster and user-triggered scaledown by removing pool spec belongs to node2 now admission server shouldn't reject the request saying not able to fetch the node using node selector(Without this PR this test case will fail).

mittachaitu: Moreover here it is component `UnitTest` case.

pkg/webhook/cspc_test.go
```diff
@@ -608,8 +620,79 @@ func TestValidateCSPCUpdateRequest(t *testing.T) {
 					},
 				},
 			},
-			expectedRsp: false,
-			getCSPCObj:  getCSPCObject,
+			blockdevicePrefix: "blockdevice",
+			expectedRsp:       false,
+			getCSPCObj:        getCSPCObject,
+		},
+		"Positive stripe expansion test when node name chaged": {
```
prateekpandey14, line 627: chaged -> changed ?

PR #138, opened 2020-07-27
chore(docs): add troubleshooting steps to migrate cStor volume from lost pool to new pool

mittachaitu:
This PR adds a troubleshooting document to migrate/rebuild the cStor volumes from lost
pool to the new or existing cStor pool.

Signed-off-by: mittachaitu <sai.chaithanya@mayadata.io>

PR #139, opened 2020-07-28
docs(bad_disk): add troubleshooting document to recover stuck pool pod

sonasingh46:
Add troubleshooting document for recovering from stuck pool pod.

### DIFF
--- a/docs/troubleshooting/pool_operations_hung.md
+++ b/docs/troubleshooting/pool_operations_hung.md
@@ -5,7 +5,7 @@ restart.
 
 Pool is always imported before creation. For the first time, when the pool creation happens, all of the 
 devices are scanned and as there is not a existing pool, a fresh pool is created. Now, when the pool is created the 
-particiapting devices are chached for faster import of the pool (in case the pool manager pod restarts.) Now if the 
+particiapting devices are chached for faster import of the pool (in case the pool manager pod restarts). Now if the 
 import happens by using the cache then this problem will not occur but there is a chance of import without cache (when 
 pool is being created for the first time)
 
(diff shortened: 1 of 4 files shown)


docs/troubleshooting/pool_operations_hung.md
```diff
@@ -0,0 +1,25 @@
...
+
+# How to recover from this situation?
+
+Now there could be two different situations :
+
+1. The device that has gone bad is actually a part of the cStor pool on the node. If this is the case
+then see [Block Device Replacement](../tutorial/cspc/tuning/tune.md) in this tutorial. This tutorial is for
+`mirror` RAID but similar approach can be followed for other RAID configurations like `raidz1` and `raidz2`. 
+
+**Note:** Block device replacement is not supported for `stripe` raid configuration.
```
mittachaitu, line 20: Q: For stripe case do we need to point to https://github.com/openebs/cstor-operators/pull/138? This means in case of stripe pool if the disk is lost/corrupted then we need to remove pool spec and create new pool spec and scale volume replicas in that pool.

docs/troubleshooting/pool_operations_hung.md
```diff
@@ -0,0 +1,34 @@
+# Pool Operation Hung Due to Bad Disk
+
+As of now, cStor scans all the devices on the node when it tries to import the pool in case of a pool manager pod
+restart.
+
+Pool is always imported before creation. For the first time, when the pool creation happens, all of the 
+devices are scanned and as there is not a existing pool, a fresh pool is created. Now, when the pool is created the 
+particiapting devices are chached for faster import of the pool (in case the pool manager pod restarts.) Now if the
```
mittachaitu, line 8: nit:
```suggestion
particiapting devices are chached for faster import of the pool (in case the pool manager pod restarts). Now if the 
```

PR #119, opened 2020-07-07
chore(operator): add generic csi-operator yaml

prateekpandey14:
**Why we need these changes:**

To support CStor CSI drivers across all the linux distribution, as we have dependencies to mount iscsi libs path into the csi node daemon sets, which is going to be different on the bases of Linux distro (like Ubuntu, CentOS, CoreOS, SuSe, Flatcar, etc).
 
Most modern Linux distros use the open-iscsi project to build their iSCSI packages. Different distros use different versions of open-iscsi. Some versions of open-iscsi have additional library dependencies.

*Options:*
- first solution is to maintain different yaml for each distribution with specific lib mounts and keep them updated. which is difficult to maintain in the long run and keep them updated.
- Run container bind-mounted with host root "/" filesystem 

Here are some links of different drivers and blogs advice to mount the host root :
1. https://www.docker.com/blog/road-to-containing-iscsi/
2. https://github.com/hpe-storage/co-deployments/blob/11682c3fb400998b8bd852ceff4e5d1a7943f7a4/helm/charts/hpe-csi-driver/templates/hpe-csi-node.yaml#L124
3. https://github.com/IBM/ibm-block-csi-operator/blob/7eb62a2b2bc144bb86554d37123c4ecb28c2c145/pkg/controller/ibmblockcsi/syncer/csi_node.go#L310

**Mount Point Changes**
- Earlier Volume Mounts
 
```yaml
        - name: iscsiadm-bin
          hostPath:
            path: /sbin/iscsiadm
            type: File
        - name: iscsiadm-lib-isns-nocrypto
          hostPath:
            path: /lib/x86_64-linux-gnu/libisns-nocrypto.so.0
            type: File
```
- With new changes in volume mounts:

```yaml
        - name: chroot-iscsiadm
          configMap:
            defaultMode: 0755
            name: openebs-cstor-csi-iscsiadm      // mentioned below
        - name: host-root
          hostPath:
            path: /
            type: Directory
```
### Config map to find out the `iscsiadm` bin path, the chroot command example is 

```bash
chroot /jail /path/to/some/command <parameters>
```

```yaml

kind: ConfigMap
apiVersion: v1
metadata:
  name: openebs-cstor-csi-iscsiadm
  namespace: kube-system
data:
  iscsiadm: |
    #!/bin/sh
    if [ -x /host/sbin/iscsiadm ]; then        // Make sure the /host/sbin/iscsiadm is statically linked and executable
[...]

PR #140, opened 2020-07-28
chore(yaml): add generic CSI operator yaml with root mount

prateekpandey14:

As a container process whenever any other pods get Restart/Delete, container has to umount the specific mounted paths and has to be unlinked any reference mounts from the host as well.
Because the CSI Node daemonsets mount the `/:/host` (host root),  and the daemonset container can see the other mounts from the host as well.
So when any pod gets restart/delete all the other mount related info has to be propagated to the CSI node container, if not then pod will be stuck in `Terminated` state due to error "device or resource busy" while unlink the mounts 

But there is different behavior has been observed against the containerd versions installed in k8s clusters.

Issue for reference https://github.com/containerd/containerd/issues/3667

**Konvoy Clusters With MountPropagation needs to be set explicitly**
- CentOS: 7.8 version
- Containerd version: 1.3.4
 
So when any pod gets restart/delete all the mount related info are not getting  propagated to the CSI node container, unless we explicitly set the MountPropagation, otherwise  pod will be stuck in `Terminated` state due to error "device or resource busy" while unlink the mounts 

```
failed to get sandbox ip: check network namespace closed: remove netns: unlinkat /var/run/netns/cni-413cfa2b-b915-7561-cbf3-1e24f1093e85: device or resource busy
```

Below are the error outputs that show the mounts can be seen and holded by a CSI Node process:

```sh
[root@konvoy-node5 ~]# grep cni-413cfa2b-b915-7561-cbf3-1e24f1093e85 /proc/*/mountinfo
/proc/28724/mountinfo:1853 1813 0:3 / /host/run/netns/cni-413cfa2b-b915-7561-cbf3-1e24f1093e85 rw,nosuid,nodev,noexec,relatime - proc proc rw

$ ps -aux | grep 28724
root     28724  0.0  0.0 490676 25548 ?        Ssl  20:53   0:00 /usr/local/bin/cstor-csi-driver --nodeid=konvoy-node5.mayalabs.io --endpoint=unix:///plugin/csi.sock --url=https://openebs.io --plugin=node
root     31892  0.0  0.0 112712   960 pts/1    S+   20:56   0:00 grep --color=auto 28724
```

**Followed below steps to reproduce the issue:**
```yaml
          volumeMounts:
            - name: host-root
              mountPath: /host
            - name: host-test
              mountPath: /test
     volumes:
        - name: host-root
          hostPath:
            path: /
            type: Directory
        - name: host-test
          hostPath:
            path: /root/test
            type: DirectoryOrCreate
```

1: check the initial mounts inside the container
```
kubectl exec -it openebs-cstor-csi-node-twtj6 -c openebs-csi-plugin -n openebs bash
$ mount | grep pawan
/dev/mapper/centos-root on /pawan type xfs (rw,relatime,seclabel,attr2,inode64,noquota)
```

2: Mounted the loop device in test/vol on Host
```
$ mkdir test/vol

$ sudo mount /dev/loop0 test/vol

$ mount | grep test
/dev/loop0 on /root/test/vol type ext4 (rw,relatime,seclabel,data=ordered)
 ```

3: Restart the pod to have updated mount paths
```
root@konvoy-master ~]# kubectl exec -it openebs-cstor-csi-node-4tdwb -c openebs-csi-plugin -n openebs bash
root@konvoy-node3:/# mount | grep test
/dev/mapper/centos-root on /test type xfs (rw,relatime,seclabel,attr2,inode64,noquota)
/dev/loop0 on /test/vol type ext4 (rw,relatime,seclabel,data=ordered)
```

5: Unmount the device from HOST and Check the mounts

```
$ sudo umount /dev/loop0

$ grep test /proc/*/mountinfo
/proc/5874/mountinfo:1030 1009 253:0 /root/test /test rw,relatime - xfs /dev/mapper/centos-root rw,seclabel,attr2,inode64,noquota
/proc/5874/mountinfo:1031 1030 7:0 / /test/vol rw,relatime - ext4 /dev/loop0 rw,seclabel,data=ordered
/proc/5999/mountinfo:1030 1009 253:0 /root/test /test rw,relatime - xfs /dev/mapper/centos-root rw,seclabel,attr2,inode64,noquota
/proc/5999/mountinfo:1031 1030 7:0 / /test/vol rw,relatime - ext4 /dev/loop0 rw,seclabel,data=ordered 
```

7: Mounts exists inside the container ( which is expected behavior as MountPropagation not set, but in older containerd version list gets updated without MountPropagation set)
```
root@konvoy-node3:/# mount | grep test
/dev/mapper/centos-root on /test type xfs (rw,relatime,seclabel,attr2,inode64,noquota)
/dev/loop0 on /test/vol type ext4 (rw,relatime,seclabel,data=ordered)
```

8. Try to remove the unmounted path ( in case of older containerd version we able to remove path successfully)
```
$ sudo rm -rf test/vol
rm: cannot remove 'test/vol': Device or resource busy
```

### Other Cluster without MountPropagation Set
- CentOS 7 with containerd version: 1.2.13
- Local Cluster containerd version: 1.2.13 

The issue was not reproducible in older version of containerd ( 1.2.13)  as the mounts are getting propagated inside the container once they get umounted from the Host, without explicitly set the MountPropagation. Due to this behavior, we were not hitting the above issue.

Mounted a test Directory inside the pod,  and followed the below steps to identify the behavior.  

1: mount list inside the csi-node container
```sh
root@centos-worker-1:/# mount | grep test
   /dev/sda2 on /test type xfs (rw,relatime,seclabel,attr2,inode64,noquota)
```

2: Created a loop device in worker-1 and mount on /var/test
```   
   $ sudo mount /dev/loop0 /var/test
```

3: check the update mount list ( mount list getting updated without any MountPropagation set)
```
$ root@centos-worker-1:/# mount | grep test
  /dev/sda2 on /test type xfs (rw,relatime,seclabel,attr2,inode64,noquota)
  /dev/loop0 on /host/var/test type ext4 (rw,relatime,seclabel,data=ordered)
```
4: umount the dir on worker-1
```
$ sudo umount /var/test
```

5: Mount removed from the list 
```
$ root@centos-worker-1:/# mount | grep test
  /dev/sda2 on /test type xfs (rw,relatime,seclabel,attr2,inode64,noquota)
```

**Solution:**
As the above issue refers the unmount has not been propagated to the container, to solve this we need to share the mounts from host to container using "HostToContainer" propagation.

```yaml
          volumeMounts: 
            - name: host-root
              mountPath: /host
              mountPropagation: "HostToContainer"
```

More info about  can be found in [PR](https://github.com/openebs/cstor-operators/pull/119) about the different drivers and blogs advices or does mount the host root :


deploy/csi-operator.yaml
```diff
@@ -962,8 +977,12 @@ spec:
               # needed so that any mounts setup inside this container are
               # propagated back to the host machine.
               mountPropagation: "Bidirectional"
-            - name: iscsiadm-bin
+            - name: host-root
+              mountPath: /host
+              mountPropagation: "HostToContainer"
+            - name: chroot-iscsiadm
               mountPath: /sbin/iscsiadm
+              subPath: iscsiadm
```
mittachaitu, line 985: Q: Why we need subpath?

prateekpandey14: because we don't want to mount as volume , i.e. to not change the path inside the container

prateekpandey14: This also ensures that any symlinks get resolved to inside the chroot environment.

deploy/csi-operator.yaml
```diff
@@ -861,6 +855,27 @@ roleRef:
...
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: openebs-cstor-csi-iscsiadm
+  namespace: kube-system
+data:
+  iscsiadm: |
+    #!/bin/sh
+    if [ -x /host/sbin/iscsiadm ]; then
+      chroot /host /sbin/iscsiadm "$@"
```
pawanpraka1, line 868: if PATH env is set, then we don't need to check for each path to see where binary is present (the last else part covers that).

deploy/csi-operator.yaml
```diff
@@ -981,7 +1000,12 @@ spec:
           hostPath:
             path: /var/lib/kubelet/
             type: Directory
-        - name: iscsiadm-bin
+        - name: chroot-iscsiadm
+          configMap:
+            defaultMode: 0755
```
pawanpraka1, line 1005: why not 0555? we just need read and execute permission, do we need write also?

deploy/csi-operator.yaml
```diff
@@ -981,7 +1000,12 @@ spec:
...
+            name: openebs-cstor-csi-iscsiadm
+        - name: host-root
           hostPath:
-            path: /sbin/iscsiadm
-            type: File
+            path: /
```
pawanpraka1, line 1009: mounting the root not in readonly mode seems to be little risky and it is the responsibility of the application running inside the conaitner, that it is not messing up with the data and corrupting anything.

deploy/csi-operator.yaml
mittachaitu: I observed iscsiadm creating required files in `/etc/iscsi/nodes` and `/etc/iscsi/send_targets`. So I think it needs permission.

pawanpraka1: @mittachaitu this is the iscsiadm script that will be mounted inside the conatiner via configmap. We just need read and execute permission for a script to execute.

mittachaitu: Ah got it(We can change to 555).

deploy/csi-operator.yaml
prateekpandey14: There can be different bin paths present based on the distribution, and even in some cases, if user installed and PATH env is not set then there is problem right?

deploy/csi-operator.yaml
prateekpandey14: Yes 555 should work fine ... let me try that once .

deploy/csi-operator.yaml
```diff
@@ -861,6 +855,27 @@ roleRef:
   name: openebs-cstor-csi-registrar-role
   apiGroup: rbac.authorization.k8s.io
 
+---
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: openebs-cstor-csi-iscsiadm
+  namespace: kube-system
```
kmova, line 863: Change to openebs-namespace and set the minimum version to 1.17

deploy/csi-operator.yaml
```diff
@@ -497,14 +491,29 @@ status:
...
+  name: cstor.csi.openebs.io
+spec:
+  # Supports persistent inline volumes.
+  volumeLifecycleModes:
+  - Persistent
+  # To determine at runtime which mode a volume uses, pod info and its
+  # "csi.storage.k8s.io/ephemeral" entry are needed.
+  podInfoOnMount: true
+  attachRequired: true
+---
```
prateekpandey14, line 507: @kmova added the CSIDriver volumeLifecycle Mode as well since we are setting up k8s 1.17.x as minimum version

akhilerm: @prateekpandey14  Does this mean the minimum version of k8s for openebs from 1.12 release should be 1.17?

deploy/csi-operator.yaml
```diff
@@ -497,14 +491,29 @@ status:
...
+
 kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: openebs-cstor-csi-snapshotter-binding
 subjects:
   - kind: ServiceAccount
     name: openebs-cstor-csi-controller-sa
-    namespace: kube-system
+    namespace: openebs
```
akhilerm, line 516: question: if in openebs namespace, can the default service account for other openebs components be used here?

deploy/csi-operator.yaml
prateekpandey14: Not for openebs overall, its specific to CSI/CSPC with all the feature supported like migration/upgrade etc.

deploy/csi-operator.yaml
prateekpandey14: I think its good practice to have a specific service account with certain permission specific to components.

deploy/csi-operator.yaml
akhilerm: question: `PATH` will be set on the host right? Not inside container. Or does the PATH change when we do `chroot`

mittachaitu:
adding hold merge label --- just observed iscsi stale sessions even after deleting the deployment.

prateekpandey14:
@kmova PR fixes that handles the iscsi stale session https://github.com/openebs/cstor-csi/pull/108

Issue #128, opened 2020-07-13
Update CSPI status to offline when the pool pod is restarted


mittachaitu:
This issue will track the CSPI status:
- Update the CSPI status.phase to `Offline` when the pool pod is restarted.
-  Update the CSPI status.Phase to `Lost` when the process not able to import/create the pool.

mittachaitu:
As discussed in the standup call whenever pool-manager gets `sig-term` signal it will update the CSPI status of `offline`

PR #141, opened 2020-07-28
fix(status): update CSPI status when corresponding cStor pool-manager terminated

mittachaitu:
This PR updates the CSPI status to **OFFLINE** whenever
cStor pool-manager container is terminated. Pod deletion, eviction,
scale down are a few scenarios that will cause the container
to terminate.

**NOTE**: Status will be not updated in case of aburupt
the shutdown of pods/nodes.

This PR fixes: #128 

### DIFF
--- a/build/pool-manager/Dockerfile
+++ b/build/pool-manager/Dockerfile
@@ -30,5 +30,5 @@ LABEL org.label-schema.build-date=$DBUILD_DATE
 LABEL org.label-schema.vcs-url=$DBUILD_REPO_URL
 LABEL org.label-schema.url=$DBUILD_SITE_URL
 
-ENTRYPOINT entrypoint.sh
+ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
 EXPOSE 7676
--- a/build/pool-manager/entrypoint.sh
+++ b/build/pool-manager/entrypoint.sh
@@ -22,7 +22,7 @@ waitForChildProcessToFinish(){
 }
 
 rm /usr/local/bin/zrepl
-exec /usr/local/bin/pool-manager start &
+/usr/local/bin/pool-manager start &
 pool_manager_pid=$!
 
 #exec service ssh start
--- a/build/pool-manager/entrypoint.sh
+++ b/build/pool-manager/entrypoint.sh
@@ -1,8 +1,34 @@
-#!/bin/sh
+#!/bin/bash
 
 set -ex
 
+#sigterm caught SIGTERM signal and forward it to pool_manager process
+_sigterm() {
+  echo "[entrypoint.sh] caught SIGTERM signal forwarding to pid [$pool_manager_pid]."
+  kill -TERM "$pool_manager_pid" 2> /dev/null
+  waitForChildProcessToFinish
+}
+
+#sigint caught SIGINT signal and forward it to pool_manager process
+_sigint() {
+  echo "[entrypoint.sh] caught SIGINT signal forwarding to pid [$pool_manager_pid]."
+  kill -INT "$pool_manager_pid" 2> /dev/null
+  waitForChildProcessToFinish
+}
+
+#waitForChildProcessToFinish waits for pool_manager process to finish
+waitForChildProcessToFinish(){
+    while ps -p "$pool_manager_pid" > /dev/null; do sleep 1; done;
+}
+
 rm /usr/local/bin/zrepl
-exec /usr/local/bin/pool-manager start
-exec service ssh start
-exec service rsyslog start
+exec /usr/local/bin/pool-manager start &
+pool_manager_pid=$!
+
+#exec service ssh start
+#exec service rsyslog start
+
+trap '_sigint' INT
+trap '_sigterm' SIGTERM
+
+wait $pool_manager_pid
--- a/pkg/controllers/cspi-controller/handler.go
+++ b/pkg/controllers/cspi-controller/handler.go
@@ -544,6 +544,7 @@ func (c *CStorPoolInstanceController) markCSPIStatusToOffline() {
 			_, err = c.clientset.CstorV1().CStorPoolInstances(cspi.Namespace).Update(&cspi)
 			if err != nil {
 				klog.Errorf("Failed to update CSPI: %s status to %s", cspi.Name, cstor.CStorPoolStatusOffline)
+				return
 			}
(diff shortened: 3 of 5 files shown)


build/pool-manager/Dockerfile
```diff
@@ -30,5 +30,5 @@ LABEL org.label-schema.build-date=$DBUILD_DATE
 LABEL org.label-schema.vcs-url=$DBUILD_REPO_URL
 LABEL org.label-schema.url=$DBUILD_SITE_URL
 
-ENTRYPOINT entrypoint.sh
+ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
```
akhilerm, line 33: @mittachaitu Does it mean that PID 1 will be `/usr/local/bin/entrypoint.sh` instead of `/bin/sh -c /usr/local/bin/entrypoint.sh`

build/pool-manager/entrypoint.sh
```diff
@@ -1,8 +1,34 @@
...
+#waitForChildProcessToFinish waits for pool_manager process to finish
+waitForChildProcessToFinish(){
+    while ps -p "$pool_manager_pid" > /dev/null; do sleep 1; done;
+}
+
 rm /usr/local/bin/zrepl
-exec /usr/local/bin/pool-manager start
-exec service ssh start
-exec service rsyslog start
+exec /usr/local/bin/pool-manager start &
```
akhilerm, line 25: Why do we need `exec` here? Can't we directly start the pool-manager.

question: Does using exec here replaces the current process with the pool manager, or is a fork created automatically?

pkg/controllers/cspi-controller/handler.go
```diff
@@ -526,6 +527,30 @@ func (c *CStorPoolInstanceController) reconcileVersion(cspi *cstor.CStorPoolInst
...
 
+// markCSPIStatusToOffline will fetch all the CSPI resources present
+// in etcd and mark it's own CSPI.Status to Offline
+func (c *CStorPoolInstanceController) markCSPIStatusToOffline() {
+	cspiList, err := c.clientset.CstorV1().CStorPoolInstances("").List(metav1.ListOptions{})
+	if err != nil {
+		klog.Errorf("Failed to fetch CSPI list error: %v", err)
+	}
+	// Fetch CSPI UUID of this pool-manager
+	cspiID := os.Getenv(OpenEBSIOCSPIID)
```
akhilerm, line 538: `OpenEBSIOCSPIID`, can this env made a little more readable. ?

pkg/controllers/cspi-controller/handler.go
```diff
@@ -526,6 +527,30 @@ func (c *CStorPoolInstanceController) reconcileVersion(cspi *cstor.CStorPoolInst
...
+	for _, cspi := range cspiList.Items {
+		if string(cspi.GetUID()) == cspiID {
+			cspi.Status.Phase = cstor.CStorPoolStatusOffline
+			// There will be one-to-one mapping between CSPI and pool-manager
+			// So after finding good to break
+			_, err = c.clientset.CstorV1().CStorPoolInstances(cspi.Namespace).Update(&cspi)
+			if err != nil {
+				klog.Errorf("Failed to update CSPI: %s status to %s", cspi.Name, cstor.CStorPoolStatusOffline)
```
akhilerm, line 546: Need a `break` here, else will print both log messages if the update fails.

build/pool-manager/Dockerfile
mittachaitu: Yes @akhilerm

pkg/controllers/cspi-controller/handler.go
mittachaitu: Basically it is representing `openebs.io/cspi-id`. Any suggestions?

build/pool-manager/entrypoint.sh
mittachaitu: exec replaces the current process with the pool manager.

> Why do we need exec here? Can't we directly start the pool-manager.

Just following whatever exists. Let me try out without exec and observe the behavior.

mittachaitu: > Can't we directly start the pool-manager.

I will update it to run directly.

pkg/controllers/cspi-controller/handler.go
```diff
@@ -526,6 +527,31 @@ func (c *CStorPoolInstanceController) reconcileVersion(cspi *cstor.CStorPoolInst
...
+func (c *CStorPoolInstanceController) markCSPIStatusToOffline() {
+	cspiList, err := c.clientset.CstorV1().CStorPoolInstances("").List(metav1.ListOptions{})
+	if err != nil {
+		klog.Errorf("Failed to fetch CSPI list error: %v", err)
+	}
+	// Fetch CSPI UUID of this pool-manager
+	cspiID := os.Getenv(OpenEBSIOCSPIID)
+	for _, cspi := range cspiList.Items {
+		if string(cspi.GetUID()) == cspiID {
+			cspi.Status.Phase = cstor.CStorPoolStatusOffline
```
mittachaitu, line 541: This will cause a problem if the pool is not created and if the pod is restarted. Need to add a check over here.

mittachaitu: done

pkg/controllers/cspi-controller/handler.go
```diff
@@ -185,6 +186,16 @@ func (c *CStorPoolInstanceController) reconcile(key string) error {
...
 	if !cspi.IsEmptyStatus() && !cspi.IsPendingStatus() {
+		// This scenario will occur when the zpool command hung due to same/someother
+		// bad disk existence in the system (or) when the underlying pool disk is lost.
+
+		// NOTE: If zpool command hung cstor-pool container in pool manager will restart
+		// because of due liveness on the pool. If cstor-pool container is killed then
+		// zpool commands will error out and fall into this scenario
+		c.recorder.Event(cspi, corev1.EventTypeWarning,
+			string(common.FailedSynced),
+			fmt.Sprintf("Unable to import a pool may be due to underlying pool disk might be lost or bad disk might be attached to the node"),
```
mittachaitu, line 197: @sonasingh46 is there any better suggestion on event message?

sonasingh46: `Failed to import the pool as the underlying pool might be lost or the disk has gone bad`

PR #147, opened 2020-08-03
chore(validations): add check for BD tag presence

sonasingh46:
BD can be tagged in order to reserve it to be used later or by other storage
engine. If a tag is found in BD then it should not be used by cstor. This commit
adds a check for this. If a tag is found in BD and is used in CSPC for cStor
pool provisioning then the cStor webhook server will reject the request.

**NOTE:** This PR https://github.com/openebs/api/pull/62/files needs to be merged first and then a rebase will be required for the current PR.

### DIFF
--- a/pkg/webhook/cspc.go
+++ b/pkg/webhook/cspc.go
@@ -413,6 +413,23 @@ func validateBlockDevice(bd *openebsapis.BlockDevice, hostName string) error {
 			bd.Labels[types.HostNameLabelKey],
 		)
 	}
+
+	if bd.Labels[types.HostNameLabelKey] != hostName {
+		return errors.Errorf(
+			"block device %s doesn't belongs to node %s",
+			bd.Name,
+			bd.Labels[types.HostNameLabelKey],
+		)
+	}
+
+	if v, found := bd.Labels[types.BlockDeviceTagLabelKey]; found {
+		return errors.Errorf(
+			"block device %s is tagged with a value %s and cannot be used",
+			bd.Name,
+			v,
+		)
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


pkg/webhook/cspc.go
```diff
@@ -408,6 +408,23 @@ func validateBlockDevice(bd *openebsapis.BlockDevice, hostName string) error {
 			bd.Labels[types.HostNameLabelKey],
 		)
 	}
+
+	if bd.Labels[types.HostNameLabelKey] != hostName {
```
akhilerm, line 412: Is the hostname check is also part of this?

mittachaitu, line 412: This check is repeated

pkg/webhook/cspc.go
```diff
@@ -408,6 +408,23 @@ func validateBlockDevice(bd *openebsapis.BlockDevice, hostName string) error {
...
+		return errors.Errorf(
+			"block device %s doesn't belongs to node %s",
+			bd.Name,
+			bd.Labels[types.HostNameLabelKey],
+		)
+	}
+
+	if v, found := bd.Labels[types.BlockDeviceTagLabelKey]; found {
```
mittachaitu, line 420: Q: @akhilerm is it possible to have blockdevice tag value as empty?

akhilerm: No @mittachaitu . . You should specify the value also. If value is empty the BD will be totally unusable as it can't be claimed by anyone.

pkg/webhook/cspc.go
```diff
@@ -413,6 +413,23 @@ func validateBlockDevice(bd *openebsapis.BlockDevice, hostName string) error {
...
 		)
 	}
+
+	if bd.Labels[types.HostNameLabelKey] != hostName {
+		return errors.Errorf(
+			"block device %s doesn't belongs to node %s",
+			bd.Name,
+			bd.Labels[types.HostNameLabelKey],
+		)
+	}
```
prateekpandey14, line 423: We already have a similar check above ^^

sonasingh46: Yes. I think rebasing brought this issue back. Fixed it. Thanks.

PR #152, opened 2020-08-11
chore(version): update upgrade matrix with valid current versions

shubham14bajpai:

This PR adds `1.12.0` & `2.0.0` as valid current versions to upgrade matrix for upgrade support from `1.12.0` and RC upgrades in `2.0.0`.

This PR fixes the issue: https://github.com/openebs/openebs/issues/3098

### DIFF
--- a/pkg/version/util.go
+++ b/pkg/version/util.go
@@ -6,7 +6,8 @@ import (
 
 var (
 	validCurrentVersions = map[string]bool{
-		"1.9.0": true, "1.10.0": true, "1.11.0": true,
+		"1.10.0": true, "1.11.0": true, "1.12.0": true,
+		"2.0.0": true,
 	}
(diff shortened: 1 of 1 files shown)


pkg/version/util.go
```diff
@@ -6,7 +6,8 @@ import (
 
 var (
 	validCurrentVersions = map[string]bool{
-		"1.9.0": true, "1.10.0": true, "1.11.0": true,
+		"1.10.0": true, "1.11.0": true, "1.12.0": true,
```
mittachaitu, line 9: Will it be good if we automate this something like below:
```go
validCurrentVersions = func(version string) bool {
if version < currentVersion {
 return true
}
return false
}
```
We can take this next release(So that we will have more test cycles)

shubham14bajpai: Sure @mittachaitu We need to verify different scenarios before going with this change.

PR #159, opened 2020-08-17
fix(roThreshold): update default roThresholdLimit to 85% during sync time

mittachaitu:
This PR updates the cStorPoolInstance with the default
value of the ROThresholdLimit if user doesn't specify
ROThresholdLimit. The default value of ROThresholdLimit
is 85%. 

**NOTE**:
- This PR is to just display roThresholdLimit on CSPI resources(
   It is already handled internally in CSPI controller).

### DIFF
--- a/pkg/controllers/cspc-controller/sync.go
+++ b/pkg/controllers/cspc-controller/sync.go
@@ -466,6 +466,7 @@ https://github.com/openebs/api/tree/master/design/cstor/v1
 ToDo: Offload this defaulting mechanism to mutating webhook server.
 */
 func defaultPoolConfig(cspi *cstor.CStorPoolInstance, cspc *cstor.CStorPoolCluster) {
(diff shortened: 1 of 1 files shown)


pkg/controllers/cspc-controller/sync.go
```diff
@@ -466,6 +466,7 @@ https://github.com/openebs/api/tree/master/design/cstor/v1
 ToDo: Offload this defaulting mechanism to mutating webhook server.
 */
 func defaultPoolConfig(cspi *cstor.CStorPoolInstance, cspc *cstor.CStorPoolCluster) {
+	defaultROThresholdLimit := 85
```
prateekpandey14, line 469: we can move this const to the top of file with other const/var types.

mittachaitu: Done

sonasingh46 (approved):

PR #162, opened 2020-08-21
fix(write cache): create claim for write cache raid group blockdevices

mittachaitu:
This PR creates claims for write cache raid groups
and wait till they get claimed before provisioning
the pools.


pkg/webhook/cspc.go
```diff
@@ -413,7 +413,7 @@ func validateBlockDevice(bd *openebsapis.BlockDevice, hostName string) error {
 			bd.Labels[types.HostNameLabelKey],
 		)
 	}
-	
+
```
prateekpandey14, line 416: remove the extra lines

mittachaitu: Done

sonasingh46 (approved):

PR #166, opened 2020-08-25
feat(ci): auto-assign github PR reviewers and assignee using github-action

prateekpandey14:
Use a GitHub Action to automatically assign GitHub PRs to the author, as
well as add reviewers.

### DIFF
--- a/.github/auto_assign.yaml
+++ b/.github/auto_assign.yaml
@@ -1,16 +1,17 @@
 addReviewers: true
 
-addAssignees: author
+addAssignees: false
 
 # Set this to the length of the reviewers list
 # because the default was not including everyone
-numberOfReviewers: 3
+numberOfReviewers: 2
 
 reviewers:
   - prateekpandey14
-  - sonasingh46
-  - mittachaitu
   - shubham14bajpai
+  - mittachaitu
+  - sonasingh46
+  - kmova
 
 skipKeywords:
   - wip
--- a/.github/auto_assign.yaml
+++ b/.github/auto_assign.yaml
@@ -1,0 +1,16 @@
+addReviewers: true
+
+addAssignees: author
+
+# Set this to the length of the reviewers list
+# because the default was not including everyone
+numberOfReviewers: 3
+
+reviewers:
+  - prateekpandey14
+  - sonasingh46
+  - mittachaitu
+  - shubham14bajpai
+
+skipKeywords:
+  - wip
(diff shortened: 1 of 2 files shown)


.github/workflows/auto_assign_prs.yml
```diff
@@ -0,0 +1,15 @@
+name: 'Auto Assign PRs'
```
kmova, line 1: this is probably not required. just go with .github/auto_assign.yaml

.github/auto_assign.yaml
```diff
@@ -0,0 +1,16 @@
+addReviewers: true
+
+addAssignees: author
+
+# Set this to the length of the reviewers list
+# because the default was not including everyone
+numberOfReviewers: 3
```
kmova, line 7: make this 2.

.github/auto_assign.yaml
```diff
@@ -0,0 +1,16 @@
+addReviewers: true
+
+addAssignees: author
```
kmova, line 3: keep this as false.

.github/workflows/auto_assign_prs.yml
prateekpandey14: as per the docs here https://github.com/marketplace/actions/auto-assign-action they suggest to have 2 files

prateekpandey14: let me try .. if it works since the name is same .

.github/auto_assign.yaml
prateekpandey14: done

.github/workflows/auto_assign_prs.yml
prateekpandey14: No, it needs separate configuration file for the auto-assign action

kmova: It is working without the workflow here: https://github.com/openebs/Mayastor/tree/develop/.github

prateekpandey14: right , i was looking for more example like this to better mange the PRs based on the types
 here https://github.com/vmware-tanzu/velero/blob/main/.github/workflows/auto_assign_prs.yml .
 
Should we go with this way or ?

kmova: I see .. and this gives the flexibility to add auto-labeling job as well.

kmova (approved):

PR #168, opened 2020-08-26
fix(backup): fixing deletion of backup snapshot

mynktl:
Changes:
- This PR fixes snapshot deletion during backup delete request.
  Existing backup endpoint uses cstor-backup resource's namespace to fetch relevant cstorvolume resource for PV, to delete the snapshot. This will result into error since cstorvolume resource are created in operator namespace instead of application namespace.

Signed-off-by: mayank <mayank.patel@mayadata.io>


mittachaitu (approved):
changes are good

PR #172, opened 2020-09-07
feat(cspc): add capability to specify allowed BD tags on CSPC

sonasingh46:
This PR adds capability to specify allowed BD tags on CSPC via
annotations.
The annotation key is `cstor.openebs.io/allowed-bd-tags` and allowed value
is comma separated strings.
For example "fast,slow" is a possible value and in this case CSPC provisioner
will reject BDs that have a BD tag (`openebs.io/block-device-tag`) present on
it and the value is other than fast or slow.

### DIFF
--- a/pkg/cspc/algorithm/select_node.go
+++ b/pkg/cspc/algorithm/select_node.go
@@ -32,9 +32,6 @@ import (
 
 const (
 	unit = 1024
-	// CStorBDTagAnnotationKey is the annotation key for CSPC allowed BD tags
-	// ToDo: Move to API type (openebs/api)
-	OpenEBSAllowedBDTagKey = "openebs.io/allowed-bd-tags"
 )
 
 // SelectNode returns a node where pool should be created.
@@ -168,7 +165,7 @@ func (ac *Config) ClaimBD(bdObj openebsio.BlockDevice) error {
 
 	// If the BD has a BD tag present then we need to decide whether
 	// cStor can use it or not.
-	// If there is not BD tag present on BD then still the BD is safe to use.
+	// If there is no BD tag present on BD then still the BD is safe to use.
 	value, ok := bdObj.Labels[types.BlockDeviceTagLabelKey]
 	var allowedBDTags map[string]bool
 	if ok {
@@ -185,7 +182,7 @@ func (ac *Config) ClaimBD(bdObj openebsio.BlockDevice) error {
 		if !allowedBDTags[strings.TrimSpace(value)] {
 			return errors.Errorf("cannot use bd {%s} as it has tag %s but "+
 				"cspc has allowed bd tags as %s",
-				bdObj.Name, value, ac.CSPC.GetAnnotations()[OpenEBSAllowedBDTagKey])
+				bdObj.Name, value, ac.CSPC.GetAnnotations()[types.OpenEBSAllowedBDTagKey])
 		}
 	}
 
@@ -228,7 +225,7 @@ func (ac *Config) ClaimBD(bdObj openebsio.BlockDevice) error {
 // Then, a map {"fast":true,"slow":true} is returned.
 func getAllowedTagMap(cspcAnnotation map[string]string) map[string]bool {
 	allowedTagsMap := make(map[string]bool)
-	allowedTags := cspcAnnotation[OpenEBSAllowedBDTagKey]
+	allowedTags := cspcAnnotation[types.OpenEBSAllowedBDTagKey]
 	if strings.TrimSpace(allowedTags) == "" {
 		return allowedTagsMap
 	}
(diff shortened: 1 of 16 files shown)


pkg/cspc/algorithm/select_node.go
```diff
@@ -162,6 +166,29 @@ func (ac *Config) ClaimBD(bdObj openebsio.BlockDevice) error {
...
+		if strings.TrimSpace(value) == "" {
+			return errors.Errorf("failed to create block device "+
+				"claim for bd {%s} as it has empty value for bd tag", bdObj.Name)
+		}
+
+		// If the BD tag in the BD is present in allowed annotations on CSPC then
+		// it means that this BD can be considered in provisioning else it should not
+		// be considered
+		allowedBDTags = getAllowedTagMap(ac.CSPC.GetAnnotations())
+		if !allowedBDTags[strings.TrimSpace(value)] {
```
akhilerm, line 185: :+1:

pkg/webhook/cspc.go
```diff
@@ -413,12 +414,17 @@ func validateBlockDevice(bd *openebsapis.BlockDevice, hostName string) error {
...
+
+	// If the BD tag is present on BD and the value is empty then
+	// this BD is not a valid BD for provisioning.
 	if v, found := bd.Labels[types.BlockDeviceTagLabelKey]; found {
-		return errors.Errorf(
-			"block device %s is tagged with a value %s and cannot be used",
-			bd.Name,
-			v,
-		)
+		if strings.TrimSpace(v) == "" {
```
akhilerm, line 421: Should we still deny from webhook if the BD has a tag?

akhilerm (commented):
Have a question on the webhook part @sonasingh46

pkg/webhook/cspc.go
sonasingh46: I am looking into how we can get rid of the webhook for most of the validations. (Reason: We faced a timeout issue ) Also, there can be a possibility that some improvements can happen in the webhook with regard to timeouts, and if that  is  the case -- putting validation in the webhook helps in two way :
1. It is an instant failure and lookup will not be required by the admins to make the CSPC right
2. Failures if uncorrected -- can cause a lot of logs to get accumulated due to reconcile loop -- but if rejected in webhook -- this can be avoided. 

Had other validations in webhook -- so took a decision to put in there too until we finalise a solution.

sonasingh46: Was this your question?

sonasingh46: If BD has a tag and the tag has an empty value -- cStor rejects it.
cStor accepts BD that are not tagged or BD with tags that should be allowed via annotation on CSPC.

akhilerm: got it.

akhilerm (approved):
lgtm

pkg/webhook/cspc_operations.go
```diff
@@ -472,6 +501,14 @@ func (pOps *PoolOperations) ClaimBD(newBdObj *openebsapis.BlockDevice, oldBD str
 		WithCapacity(resource.MustParse(ByteCount(newBdObj.Spec.Capacity.Storage))).
 		WithCSPCOwnerReference(getBDOwnerReference(pOps.OldCSPC)[0]).
 		WithFinalizer(types.CSPCFinalizer)
+	// ToDo: Move this to openebs/api builder
```
mittachaitu, line 504: :+1:

mittachaitu (commented):
Provided comments

pkg/cspc/algorithm/select_node.go
```diff
@@ -31,6 +32,9 @@ import (
 
 const (
 	unit = 1024
+	// CStorBDTagAnnotationKey is the annotation key for CSPC allowed BD tags
+	// ToDo: Move to API type (openebs/api)
+	CStorBDTagAnnotationKey = "cstor.openebs.io/allowed-bd-tags"
```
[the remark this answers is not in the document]
sonasingh46: Makes sense,Sai. Will change it.

pkg/cspc/algorithm/select_node.go
```diff
@@ -162,6 +166,29 @@ func (ac *Config) ClaimBD(bdObj openebsio.BlockDevice) error {
 		return errors.Errorf("failed to get capacity from block device %s:%s", bdObj.Name, err)
 	}
 
+	// If the BD has a BD tag present then we need to decide whether
+	// cStor can use it or not.
+	// If there is not BD tag present on BD then still the BD is safe to use.
```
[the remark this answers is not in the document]
mittachaitu: @sonasingh46 FYI typo is not yet fixed in latest commit

mittachaitu (approved):
changes are good

pkg/cspc/algorithm/select_node.go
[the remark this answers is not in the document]
sonasingh46: done. fixed it.

mittachaitu (approved):
changes are good

mittachaitu (approved):
changes are good

PR #185, opened 2020-09-23
chore(ci): set up helm chart ci

sonasingh46:
This PR adds CI for cStor helm chart linting and release.

**Note:** The cStor helm charts will reside in a directory structure: `deploy/helm`

The following are the details of changes : 
1. Adds `Chart Lint and Test` action to test helm chart changes. 
- This action is only triggered if there is any change in the content of `deploy/helm` directory.

2. Modifies the existing `Go` action.
- As a result of the modification, the action will not run for pull requests if there are no changes outside of the `deploy/helm` directory.

3. Adds a `Release Charts` action to release the helm chart. 
- This action triggers only when the push in against master and there is a change in the content of `deploy/helm` directory.


.github/workflows/lint-test.yaml
```diff
@@ -0,0 +1,39 @@
+name: Lint and Test Charts
+
+on:
+  push:
+    paths:
+      - 'charts/**'
```
kmova, line 6: can this chart code be moved to `deploy/charts`, which might require updates to github actions or customization. Pushing to `deploy/charts` will help consolidate all the deployment code under one directory.

.github/workflows/go.yml
```diff
@@ -1,5 +1,9 @@
 name: Go
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+    paths-ignore:
+      - 'deploy/helm/**'
```
akhilerm, line 6: :+1:

akhilerm (approved):
/lgtm

PR #197, opened 2020-11-02
refact(crds): update crds to use apiextensions.k8s.io/v1 apis

prateekpandey14:
- Bump NDM version 0.9.1
- Update crds to use v1 version

### DIFF
--- a/deploy/crds/all_cstor_crds.yaml
+++ b/deploy/crds/all_cstor_crds.yaml
@@ -1,26 +1,13 @@
 
 ---
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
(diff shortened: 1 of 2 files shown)


shubham14bajpai:
@prateekpandey14 Can you add the CRD as well so that it is not missed while raising charts PR
https://github.com/openebs/upgrade/blob/master/ci/migrate/migrationCRD.yaml

prateekpandey14:
@shubham14bajpai added the migrationTask CRD. as discussed upgradeTask CRD is still in `v1beta1` since its needs some api tag level changes which needs to be conveyed in other places as well.

shubham14bajpai (approved):

sonasingh46 (approved):

PR #202, opened 2020-11-11
refact(build): run only amd64 and arm64 builds

shubham14bajpai:

Enabling more than two arch in the multi-arch builds
is resulting in increased build times as well as
intermittent failures in running the builds.

Limiting to two archs - amd64 and arm64 for now.


prateekpandey14 (approved):
lgtm

PR #204, opened 2020-11-12
fix(build): propagate RELEASE_TAG to buildx container

shubham14bajpai:

The RELEASE_TAG env set in the action is not propagated to the buildx container. This PR fixes the same.
This PR also removes the deprecated envs


prateekpandey14 (approved):
lgtm

PR #210, opened 2020-11-30
fix(cspc): cleanup pending bdcs on cspc deletion

shubham14bajpai:

**What this PR does:**
This PR fixes the cleanup issue of BDCs when the CSPIs doesn't get provisioned and CSPC is deleted. 

**How to replicate the scenario?**
Setup the cstor-operator yaml and check whether the BDs are active. Then scale down the ndm-operator deployment, this will prevent the BDCs from getting bound. If we delete the CPSC now the BDCs get the deletion timestamp but the cspc-operator does not remove the finalizer from the pending BDCs.

Tested the below code in the same scenario and the BDCs got cleaned up successfully.

### DIFF
--- a/pkg/controllers/cspc-controller/sync.go
+++ b/pkg/controllers/cspc-controller/sync.go
@@ -270,7 +270,9 @@ func (c *Controller) removeCSPCFinalizer(cspc *cstor.CStorPoolCluster) error {
 			LabelSelector: string(types.CStorPoolClusterLabelKey) + "=" + cspc.Name,
 		},
 	)
-
+	if err != nil {
+		return err
+	}
 	for _, bdcItem := range bdcList.Items {
 		bdcItem := bdcItem // pin it
 		bdcObj := &bdcItem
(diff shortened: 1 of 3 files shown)


pkg/controllers/cspc-controller/sync.go
```diff
@@ -261,6 +262,25 @@ func (c *Controller) removeCSPCFinalizer(cspc *cstor.CStorPoolCluster) error {
...
 	}
 
+	// If the BD(s) are claimed and the CSPI(s) do not spawn up for some reason,
+	// the pending BDC(s) finalizer don't get removed automatically.
+	// The below code handles cleanup for such cases.
+	bdcList, err := c.GetStoredOpenebsVersionClient().BlockDeviceClaims(cspc.Namespace).List(
+		metav1.ListOptions{
+			LabelSelector: string(types.CStorPoolClusterLabelKey) + "=" + cspc.Name,
+		},
+	)
```
mittachaitu, line 272: Can we handle the error? else it will be the same behavior(BDC will exist) when some error occurs

shubham14bajpai: Done

mittachaitu (approved):
changes are good

prateekpandey14 (approved):
lgtm

PR #218, opened 2020-12-30
chore(operator): bump k8s csi to latest stable container images

prateekpandey14:
**What this PR does**:

- Bump k8s csi sidecar container image to latest stable  releases minimum kubernetes version to support 1.17

-  Remove the deprecated Cluster Driver Registrar container which is no longer required since the `CSIDriver` object instance should be created by CSI driver installer (or manually)

- Update the Snapshot specific CRDs to `V1` version

- Remove the `--provisioner` flag not required (not present)
  in new version of csi-provisioner

- Added the `--default-fstype` flag required to be set new
  csi-provisioner release

### DIFF
--- a/deploy/csi-operator.yaml
+++ b/deploy/csi-operator.yaml
@@ -13,45 +13,6 @@
 ###########  CSI Node and Driver CRDs   ############
 ###########                             ############
 ####################################################
-
-apiVersion: apiextensions.k8s.io/v1beta1
-kind: CustomResourceDefinition
-metadata:
-  creationTimestamp: null
-  name: csinodeinfos.csi.storage.k8s.io
-spec:
-  group: csi.storage.k8s.io
-  names:
-    kind: CSINodeInfo
-    plural: csinodeinfos
-  scope: Cluster
-  validation:
-    openAPIV3Schema:
-      properties:
-        csiDrivers:
-          description: List of CSI drivers running on the node and their properties.
-          items:
-            properties:
-              driver:
-                description: The CSI driver that this object refers to.
-                type: string
-              nodeID:
-                description: The node from the driver point of view.
-                type: string
-              topologyKeys:
-                description: List of keys supported by the driver.
-                items:
-                  type: string
-                type: array
-          type: array
-  version: v1alpha1
-status:
-  acceptedNames:
-    kind: ""
-    plural: ""
-  conditions: []
-  storedVersions: []
-
 ---
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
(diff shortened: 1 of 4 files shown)


deploy/snapshot_class.yaml
```diff
@@ -1,8 +1,7 @@
 kind: VolumeSnapshotClass
-apiVersion: snapshot.storage.k8s.io/v1beta1
+apiVersion: snapshot.storage.k8s.io/v1
 metadata:
   name: csi-cstor-snapshotclass
-  namespace: openebs
```
akhilerm, line 5: Is this a cluster scoped resource?

akhilerm: Also, I couldn't find the `VolumeSnapshotClass` in the cstor-operator.yaml in charts

deploy/csi-operator.yaml
```diff
@@ -13,45 +13,6 @@
 ###########  CSI Node and Driver CRDs   ############
 ###########                             ############
 ####################################################
-
-apiVersion: apiextensions.k8s.io/v1beta1
```
akhilerm, line 17: Is this CRD no longer required?

akhilerm (commented):
given some comments.

deploy/snapshot_class.yaml
prateekpandey14: yes its a cluster scope

deploy/csi-operator.yaml
prateekpandey14: No  , this resource is deprecated in k8s against the CSINode resource.

deploy/snapshot_class.yaml
prateekpandey14: yes ,i just checked snapshotclass file its not there, probably i missed to add that file while refactoring some of the changes.
I will add that in this release charts.

payes (approved):

deploy/snapshot_class.yaml
akhilerm: It should be added to the cstor-operators in charts repo also right?

prateekpandey14: yes

deploy/csi-operator.yaml
```diff
@@ -88,67 +51,107 @@ spec:
...
+            additionalProperties:
+              type: string
+            description: parameters is a key-value map with storage driver specific parameters for creating snapshots. These values are opaque to Kubernetes.
+            type: object
+        required:
+        - deletionPolicy
+        - driver
+        type: object
+    served: true
+    storage: false
```
akhilerm, line 96: Why is the v1 version not the storage version?

PR #220, opened 2021-01-05
Update tune.md

survivant:
typo


survivant:
Another question.  the compression is set to : 
compression: lz

which is not in the previous list.  Probably another typo to fix

kmova (approved):

PR #222, opened 2021-01-06
fix(travis): docker login to increase rate limit

shubham14bajpai:

With the rate limits on the ubuntu and alpine images, the travis is failing. Login into docker before the travis build so that the rate limit doubles on the instances.


shubham14bajpai:
The issue got resolved with the use of github-actions.

PR #223, opened 2021-01-06
chore(actions): move travis tests to github actions

shubham14bajpai:

This PR moves the sanity tests from travis to github actions by deploying a minikube using  https://github.com/marketplace/actions/setup-minikube-kubernetes-cluster

Got the above action from https://github.com/eclipse/jkube/blob/master/.github/workflows/e2e-tests.yml


.github/workflows/build.yml
```diff
@@ -31,9 +31,43 @@ jobs:
           pattern: '*.sh'
           exclude: './.git/*,./vendor/*'
 
+  e2e-tests:
+    needs: ['lint']
```
akhilerm, line 35: `lint` can run in parallel. It should not be a dependent for `e2e-tests`

.github/workflows/build.yml
```diff
@@ -31,9 +31,66 @@ jobs:
...
+  e2e-tests:
+    needs: ['unit-tests']
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        kubernetes: [v1.18.6]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2.3.4
```
akhilerm, line 66: Ideally we should stick to one version of actions/checkout. either `v2.3.4` or `v2`

.github/workflows/build.yml
```diff
@@ -31,9 +31,66 @@ jobs:
...
+      run: make license-check
+
+    - name: verify dependencies
+      run: make deps
+
+    - name: verify tests
+      run: make test
+
+  e2e-tests:
+    needs: ['unit-tests']
```
akhilerm, line 58: @shubham14bajpai Do we need e2e to be dependent on unit-tests. ? 

IMHO, `lint`, `unit-tests` and `e2e-tests` can be run in parallel. Only the final build/push needs to be dependent on all these 3 jobs.

.github/workflows/build.yml
```diff
@@ -31,9 +31,66 @@ jobs:
...
+          BRANCH="${GITHUB_REF##*/}"
+          CI_TAG=${BRANCH#v}-ci
+          if [ ${BRANCH} = "master" ]; then
+            CI_TAG="ci"
+          fi
+          echo "TAG=${CI_TAG}" >> $GITHUB_ENV
+          echo "BRANCH=${BRANCH}" >> $GITHUB_ENV
+      
+      - name: Build images locally
+        run: make all.amd64 || exit 1;
```
akhilerm, line 86: Can we use the buildx target itself, with `IMG_RESULT=load`

akhilerm (commented):
@shubham14bajpai Given some comments.

.github/workflows/build.yml
shubham14bajpai: Right now there no single target to build all images using buildx. Also the test yamls use images with suffix `-amd64`. I will take these changes in separate PR.

.github/workflows/build.yml
shubham14bajpai: I think checking the modules using make deps and running unit tests to check for syntactical errors is good before building and running the images.

akhilerm (approved):
LGTM

kmova (approved):

PR #224, opened 2021-01-06
Update policy.md

survivant:
set a example to patch CVC resource


prateekpandey14 (approved):
lgtm, thanks @survivant

Issue #237, opened 2021-01-26
admission-webhook.cstor.openebs.io validation of compression arguments in CStorPoolCluster


coboluxx:
The admission webhook expects values that are invalid for the ZFS compression setting.

**Example:**
Create a new CStorPoolCluster
```
apiVersion: cstor.openebs.io/v1
kind: CStorPoolCluster
metadata:
  name: cstor-hdd-x1-gli1-pool
  namespace: openebs
spec:
  pools:
  - dataRaidGroups:
    - blockDevices:
      - blockDeviceName: blockdevice-c89ddb7adce61f8d21b6e27a488b1940
    nodeSelector:
      kubernetes.io/hostname: iuck8s1.core.idnt.net
    poolConfig:
      dataRaidGroupType: stripe
      compression: lz4 # <- lz4 ./ lz 
```

**Result**
Failed CStorPoolInstance:
Failed to create pool due to 'Failed to create pool {cstor-6c7f1f4b-9466-4239-8743-cf0cea515df9} : Failed to create pool.. cannot create 'cstor-6c7f1f4b-9466-4239-8743-cf0cea515df9': 'compression' must be one of 'on | off | lzjb | gzip | gzip-[1-9] | zle | lz4'

lz4 or on gets rejected by the webhook:
Error from server (BadRequest): error when creating "STDIN": admission webhook "admission-webhook.cstor.openebs.io" denied the request: invalid cspc specification: invalid pool spec: unsupported compression 'lz4'
Error from server (BadRequest): error when creating "STDIN": admission webhook "admission-webhook.cstor.openebs.io" denied the request: invalid cspc specification: invalid pool spec: unsupported compression 'on' specified

prateekpandey14:
closing .. fixed in #241

PR #241, opened 2021-02-01
fix(webhook): allow cspc validation based on supported compression algos

prateekpandey14:

fixes:#237

### DIFF
--- a/pkg/webhook/cspc.go
+++ b/pkg/webhook/cspc.go
@@ -338,7 +338,7 @@ var (
 		cstor.PoolRaidz2:   isRaidz2BDCountValid,
 	}
 	// SupportedCompression is a map holding the supported compressions
-	// TODO: support to allow gzip-[1-9] compression algo can be added later if requested
+	// TODO: support to allow gzip-[1-9] compression algorithm also can be added later if requested
 	// by the community.
 	SupportedCompression = map[string]bool{
 		"on":   true,
--- a/pkg/webhook/cspc.go
+++ b/pkg/webhook/cspc.go
@@ -338,12 +338,16 @@ var (
 		cstor.PoolRaidz2:   isRaidz2BDCountValid,
 	}
 	// SupportedCompression is a map holding the supported compressions
-	// TODO: confirm all the compression types supported by control plane
-	// and update the map accordingly
+	// TODO: support to allow gzip-[1-9] compression algo can be added later if requested
+	// by the community.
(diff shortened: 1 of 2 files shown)


shubham14bajpai (approved):
lgtm

pkg/webhook/cspc.go
```diff
@@ -341,9 +341,13 @@ var (
 	// TODO: confirm all the compression types supported by control plane
 	// and update the map accordingly
 	SupportedCompression = map[string]bool{
-		"":    true,
-		"off": true,
-		"lz":  true,
+		"on":   true,
+		"off":  true,
+		"lzjb": true,
+		"gzip": true,
```
mittachaitu, line 347: We have to support values of `gzip-[1-9]` also since the data plane supports the same

mittachaitu: Command to create pool will be `zpool create <pool_name> <disk_path> -O compresstion=gzip-9`

pkg/webhook/cspc.go
```diff
@@ -341,9 +341,13 @@ var (
 	// TODO: confirm all the compression types supported by control plane
 	// and update the map accordingly
 	SupportedCompression = map[string]bool{
```
mittachaitu, line 343: Add a comment saying this can be relooked when user request for gzip-[1-9] compression support...

mittachaitu (approved):
changes are good

pkg/webhook/cspc.go
prateekpandey14: added the comment

pkg/webhook/cspc.go
```diff
@@ -338,12 +338,16 @@ var (
 		cstor.PoolRaidz2:   isRaidz2BDCountValid,
 	}
 	// SupportedCompression is a map holding the supported compressions
-	// TODO: confirm all the compression types supported by control plane
-	// and update the map accordingly
+	// TODO: support to allow gzip-[1-9] compression algo can be added later if requested
```
mittachaitu, line 341:
```suggestion
	// TODO: support to allow gzip-[1-9] compression also can be added later if requested
```

mittachaitu (approved):
changes are good

pkg/webhook/cspc.go
prateekpandey14: changed to  -- > `algorithm also`

PR #246, opened 2021-02-10
fix(webhook): cleanup validatingwebhook config on openebs namespace deletion

prateekpandey14:
- Add the openebs namespace delete validation checks
- Handle the `validatingwebhookconfiguration` garbage collection  issue due to invalid cross-namespace ownerReference checks introduced in k8s 1.20+

Refer:openebs/openebs#3338

from k8s 1.20+ onwards Cluster-scoped dependents can only specify cluster-scoped owners. In v1.20+, if a cluster-scoped dependent specifies a namespaced kind as an owner, it is treated as having an unresolveable owner reference, and is not able to be garbage collected.

In v1.20+, if the garbage collector detects an invalid cross-namespace ownerReference, or a cluster-scoped dependent with an ownerReference referencing a namespaced kind, a warning Event with a reason of OwnerRefInvalidNamespace and an involvedObject of the invalid dependent is reported. 

You can check for that kind of Event by running
```
kubectl get events -A --field-selector=reason=OwnerRefInvalidNamespace.
```

**Note:**  We have to think about how to handle the deletion if only cstor specific components has been removed from the cluster in openebs namespace, or may be webhook deployment is scaled down due to some reason. 

### DIFF
--- a/pkg/webhook/configuration.go
+++ b/pkg/webhook/configuration.go
@@ -73,13 +73,17 @@ var (
 	Ignore = admissionregistration.Ignore
 	// Fail means that an error calling the webhook causes the admission to fail.
 	Fail = admissionregistration.Fail
+	// SideEffectClassNone means that calling the webhook will have no side effects.
+	SideEffectClassNone = v1beta1.SideEffectClassNone
 	// WebhookFailurePolicye represents failure policy env name to make it configurable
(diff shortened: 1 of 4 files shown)


shubham14bajpai (approved):
lgtm

PR #259, opened 2021-03-02
fix(ci): increase timeout values for pool checks and update ndm operator

prateekpandey14:

### DIFF
--- a/deploy/ndm-operator.yaml
+++ b/deploy/ndm-operator.yaml
@@ -1,4 +1,3 @@
----
 # This is the node-disk-manager related config.
 # It can be used to customize the disks probes and filters
 apiVersion: v1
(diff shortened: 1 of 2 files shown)


deploy/ndm-operator.yaml
```diff
@@ -37,7 +36,7 @@ data:
         name: path filter
         state: true
         include: ""
-        exclude: "loop,/dev/fd0,/dev/sr0,/dev/ram,/dev/dm-,/dev/md"
+        exclude: "/dev/loop,/dev/fd0,/dev/sr0,/dev/ram,/dev/dm-,/dev/md,/dev/rbd,/dev/zd"
```
shubham14bajpai, line 39: Instead of this can we make this change in sanity.sh which is more generic to the environments
```sh
bdname=$(kubectl -n openebs get blockdevices -o jsonpath='{.items[?(@.spec.details.deviceType=="sparse")].metadata.name}')
```

prateekpandey14: Added this as well

shubham14bajpai (approved):
lgtm

PR #268, opened 2021-03-24
chore(docs): making the sample CSPC names the same

niladrih:
Making the sample CSPC name the same across examples and quickstart, so that the default storageClass can be used out of the box no matter which sample CSPC a user decides to pick up.

Ref: [Slack thread](https://kubernetes.slack.com/archives/CUAKPFU78/p1616598454223100?thread_ts=1616433706.157900&cid=CUAKPFU78)

Signed-off-by: Niladri Halder <niladri.halder@mayadata.io>

### DIFF
--- a/docs/quick.md
+++ b/docs/quick.md
@@ -194,7 +194,7 @@ follwing steps:
 2.  Apply the modified CSPC YAML.
 
     ```bash
-    kubectl apply -f cstor-storage.yaml
+    kubectl apply -f cspc-single.yaml
     ```
 3. Check if the pool instances report their status as 'ONLINE'.
 
@@ -233,6 +233,7 @@ follwing steps:
    parameters:
      cas-type: cstor
      cstorPoolCluster: cstor-storage
+     # replicaCount should be <= no. of CSPI
      replicaCount: "3"
    ```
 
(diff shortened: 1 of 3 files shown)


mittachaitu (approved):
changes are good. Do we need to make the same changes in other tutorials?

niladrih:
@mittachaitu -- Quickstart does not go into the other tutorials, so it is unlikely that the names in those tutorials will affect users following the quickstart. 

For other tutorials, if the names are consistent for the particular tutorial then I think we are good.

niladrih:
@mittachaitu -- On second thought, I agree with you, I'll change the names for the other tutorials as well. Thanks.

docs/quick.md
```diff
@@ -233,6 +233,7 @@ follwing steps:
    parameters:
      cas-type: cstor
      cstorPoolCluster: cstor-storage
```
prateekpandey14, line 235: can we add the comment here saying , name needs to be changed accroding to the name has been used to configure cspc ?

niladrih:
@mittachaitu -- Holding the changes to other tutorials. It makes sense that the `stripe` tutorial has a CSPC named `cspc-stripe`, and the `mirror` tutorial has a CSPC named `cspc-mirror-single`.

prateekpandey14 (approved):

PR #287, opened 2021-05-03
fix(helm chart): rename priority class name for CSI plugin (node and controller)

shovanmaity:
Include storage engine name in priority class name to make sure CSI driver of different storage will have different priority class. This will resolve the issue of installing multiple storage engine using helm chart.

What will happen to older priority classes(upgrade scenario)?
=> Older priority classes will be deleted and CSI node and controller driver pod will be restarted with new priority classes.

### DIFF
--- a/deploy/helm/charts/templates/priority-class.yaml
+++ b/deploy/helm/charts/templates/priority-class.yaml
@@ -1,15 +1,19 @@
+{{- if .Values.csiController.priorityClass.create }}
 apiVersion: scheduling.k8s.io/v1
 kind: PriorityClass
 metadata:
-  name: openebs-cstor-csi-controller-critical
+  name: {{ template "cstor.csiController.priorityClassName" . }}
 value: 900000000
 globalDefault: false
 description: "This priority class should be used for the CStor CSI driver controller deployment only."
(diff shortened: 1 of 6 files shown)


deploy/helm/charts/templates/priority-class.yaml
```diff
@@ -1,15 +1,15 @@
...
+  name: openebs-cstor-csi-controller-critical
 value: 900000000
 globalDefault: false
 description: "This priority class should be used for the CStor CSI driver controller deployment only."
 ---
 apiVersion: scheduling.k8s.io/v1
 kind: PriorityClass
 metadata:
-  name: openebs-csi-node-critical
+  name: openebs-cstor-csi-node-critical
```
sonasingh46, line 12: @shovanmaity -- Similar comments. See https://github.com/openebs/jiva-operator/pull/72#discussion_r625763700

sonasingh46: Again, not a blocker for merge.

sonasingh46 (approved):
Alt Text

prateekpandey14 (approved):

PR #290, opened 2021-05-04
(bugsbash)remove unused code

asquare14:
Signed-off-by: aSquare14 <atibhi.a@gmail.com>

PR #301, opened 2021-05-04
Remove unused code

soniasingla:


pkg/controllers/cstorvolumeconfig/controller_test.go
```diff
@@ -71,7 +71,6 @@ type fixture struct {
 
 	// Actions expected to happen on the client. Objects from here are also
 	// preloaded into NewSimpleFake.
-	actions        []core.Action
```
mittachaitu, line 74: Is linter shown this field as unused?

soniasingla: yes!

mittachaitu: But we are using them as we can see [here](https://github.com/openebs/cstor-operators/pull/301/checks?check_run_id=2500796432) test cases are failing.

mittachaitu: We can undo the changes done to `actions`

mittachaitu:
@soniasingla Thanks for the PR but changes in this PR are already taken into master via #290 so closing the PR.

PR #303, opened 2021-05-04
Remove unused code

soniasingla:

### DIFF
--- a/pkg/signals/signal.go
+++ b/pkg/signals/signal.go
@@ -21,14 +21,10 @@ import (
 	"os/signal"
 )
 
-var onlyOneSignalHandler = make(chan struct{})
-
 // SetupSignalHandler registered for SIGTERM and SIGINT. A stop channel is returned
 // which is closed on one of these signals. If a second signal is caught, the program
 // is terminated with exit code 1.
 func SetupSignalHandler() (stopCh <-chan struct{}) {
-	//	close(onlyOneSignalHandler) // panics when called twice
-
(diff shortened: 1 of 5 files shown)


pkg/signals/signal.go
```diff
@@ -21,8 +21,6 @@ import (
 	"os/signal"
 )
 
-var onlyOneSignalHandler = make(chan struct{})
-
```
mittachaitu, line 25: We can remove commented code related to `onlyOneSignalHandler`. Commented code exist at line number [30](https://github.com/openebs/cstor-operators/blob/84ac251371f709e1d12d7f079ea4bb274d81a7a4/pkg/signals/signal.go#L30)

pkg/webhook/webhook.go
```diff
@@ -53,7 +53,6 @@ var (
 		metav1.NamespaceSystem,
 		metav1.NamespacePublic,
 	}
-	snapshotAnnotation = "snapshot.alpha.kubernetes.io/snapshot"
```
mittachaitu, line 56: Instead of removing it, we can comment variable declaration line. We may use this variable in future. Currently, code which uses this variable was [commented](https://github.com/openebs/cstor-operators/blob/master/pkg/webhook/pvc.go#L166-L169).

mittachaitu (commented):
looks good but provided minor comments

mittachaitu:
@soniasingla Did you get a chance to take look at comments?

PR #304, opened 2021-05-04
(bugsbash) fixes superfluous nil err check before return

asquare14:
Signed-off-by: aSquare14 <atibhi.a@gmail.com>


pkg/controllers/cspc-controller/scale.go
```diff
@@ -68,10 +68,8 @@ func (pc *PoolConfig) CreateCSPI(cspc *cstor.CStorPoolCluster) error {
 		return err
 	}
 	err = pc.CreateCSPIDeployment(cspc, cspi)
```
mittachaitu, line 70: nit: Even we can do in the following way
```suggestion
      return pc.CreateCSPIDeployment(cspc, cspi)
```

mittachaitu (approved):
lgtm

asquare14:
@mittachaitu Fixed DCO, thanks for the review !

pkg/controllers/cspc-controller/scale.go
```diff
@@ -67,11 +67,9 @@ func (pc *PoolConfig) CreateCSPI(cspc *cstor.CStorPoolCluster) error {
...
 		return err
 	}
-	err = pc.CreateCSPIDeployment(cspc, cspi)
-	if err != nil {
-		return err
-	}
-	return nil
+      return pc.CreateCSPIDeployment(cspc, cspi)
+
+	return err
```
mittachaitu, line 72: We can remove this line since we are returning from line num 70 this part is unreachable
```suggestion
```

asquare14:
Thanks for the review ! Done :D

PR #307, opened 2021-05-04
Add S1039 to ignore list of muse bot

soniasingla:


pkg/server/cstorvolumeconfig/backup_endpoint.go
```diff
@@ -199,7 +199,7 @@ func (bOps *backupAPIOps) get() (interface{}, error) {
 
 	// backup name is expected
 	if len(strings.TrimSpace(backup.Spec.BackupName)) == 0 {
-		return nil, CodedError(400, fmt.Sprintf("Failed to get backup: missing backup name "))
+		return nil, CodedError(400, string("Failed to get backup: missing backup name "))
```
mittachaitu, line 202: We no need to explicitly type cast using `string`. It can be as simple like below
```suggestion
		return nil, CodedError(400, "Failed to get backup: missing backup name ")
```

pkg/volume-rpc/targetserver/config.go
```diff
@@ -45,13 +45,13 @@ func (csr *CVReplicationDetails) BuildConfigData() map[string]string {
...
-	key := fmt.Sprintf("  ReplicationFactor")
+	key := "  ReplicationFactor"
 	value := fmt.Sprintf("  ReplicationFactor %d", csr.ReplicationFactor)
 	data[key] = value
-	key = fmt.Sprintf("  ConsistencyFactor")
+	key = "  ConsistencyFactor"
 	value = fmt.Sprintf("  ConsistencyFactor %d", csr.ConsistencyFactor)
 	data[key] = value
-	key = fmt.Sprintf("  Replica %s", csr.ReplicaID)
+	key = string(csr.ReplicaID)
```
mittachaitu, line 54: It should be in the following format
```go
key = "  Replica " + csr.ReplicaID
```
above output should be **"  Replica \<replica ID\>"**(two spaces at starting of a string) but with a changes in PR it will get modified to `<replica ID>`.
Can you please update the above snippet?

mittachaitu: This is required since we are comparing the above keys over [here](https://github.com/openebs/api/blob/d18d04d8711235a9f217d20542b925f7fe8943f3/pkg/util/fileoperator.go#L74) with equality operator.

mittachaitu (changes requested):
@soniasingla Thanks for the PR, Can you PTAL at the above comments?

mittachaitu:
Hi @soniasingla, Thanks for all the PRs(#311 , #315 , #316 , #321 ) discussed with the team members related to the changes of `fmt.Sprtinf()` and felt the following points:
- go compiler will perform inlining during the compile time so these changes will be replaced by strings.
- We need to make these changes just to satisfy `gosimple` linter(which may not be good).


@soniasingla  Can you help in fixing this issue in another way to satisfy the linter?
- Just by adding rule to exclude `Unnecessary use of fmt.Sprint` errors by updating [muse config](https://github.com/openebs/cstor-operators/blob/master/.muse/config.toml#L1) file with [S1039](https://staticcheck.io/docs/checks#S1039) rule.

mittachaitu (approved):
Thanks for the PR... changes are good

kmova (approved):

PR #312, opened 2021-05-04
Remove usage of fmt.sprintf

soniasingla:


mittachaitu:
@soniasingla  Can you take look at #302 comments?  Same comments will be applicable over here.

mittachaitu (approved):
LGTM

PR #315, opened 2021-05-04
Remove usage of fmt.sprintf

soniasingla:

### DIFF
--- a/pkg/zcmd/zpool/get/builder.go
+++ b/pkg/zcmd/zpool/get/builder.go
@@ -17,7 +17,6 @@ limitations under the License.
 package pget
 
 import (
-	"fmt"
 	"os/exec"
 	"reflect"
 	"runtime"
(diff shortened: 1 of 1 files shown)


pkg/zcmd/zpool/get/builder.go
```diff
@@ -153,35 +152,35 @@ func (p *PoolGProperty) Build() (*PoolGProperty, error) {
...
-		p.appendCommand(&c, fmt.Sprintf(" -p "))
+		p.appendCommand(&c, " -p ")
 	}
 
 	if IsFieldListSet()(p) {
-		p.appendCommand(&c, fmt.Sprintf(" -o "))
+		p.appendCommand(&c, " -o ")
 		for _, f := range p.FieldList {
-			p.appendCommand(&c, fmt.Sprintf("%s,", f))
+			p.appendCommand(&c, " "+f+" ")
```
mittachaitu, line 168: This should be something like below... It is not expecting for space
```suggestion
			p.appendCommand(&c, f+", ")
```
Above is equivalent to `p.appendCommand(&c, f + ",")`

pkg/zcmd/zpool/get/builder.go
```diff
@@ -153,35 +152,35 @@ func (p *PoolGProperty) Build() (*PoolGProperty, error) {
...
 		}
 		// add space to separate next arg
-		p.appendCommand(&c, fmt.Sprintf(" "))
+		p.appendCommand(&c, " ")
 	}
 
 	if IsPropListSet()(p) {
 		for _, v := range p.PropList {
-			p.appendCommand(&c, fmt.Sprintf("%s,", v))
+			p.appendCommand(&c, " "+v+" ")
```
mittachaitu, line 176: Same as above comment

mittachaitu (commented):
Provided few comments PTAL

PR #316, opened 2021-05-04
Remove usage of fmt.sprintf

soniasingla:

### DIFF
--- a/pkg/zcmd/zpool/import/builder.go
+++ b/pkg/zcmd/zpool/import/builder.go
@@ -165,11 +165,11 @@ func (p *PoolImport) Build() (*PoolImport, error) {
 	var c strings.Builder
 	p = p.Validate()
 	p.appendCommand(&c, bin.ZPOOL)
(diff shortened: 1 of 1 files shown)


pkg/zcmd/zpool/import/builder.go
```diff
@@ -165,11 +165,11 @@ func (p *PoolImport) Build() (*PoolImport, error) {
 	var c strings.Builder
 	p = p.Validate()
 	p.appendCommand(&c, bin.ZPOOL)
-	p.appendCommand(&c, fmt.Sprintf(" %s ", Operation))
+	p.appendCommand(&c, " "+Operation+" ")
 
 	if IsPropertySet()(p) {
 		for _, v := range p.Property {
-			p.appendCommand(&c, fmt.Sprintf(" -o %s ", v))
+			p.appendCommand(&c, " "+v+" ")
```
mittachaitu, line 172: To set zpool property `-o` (option) is required but we are missing in above change

mittachaitu (commented):
Thanks for the PR... Provided comment PTAL

PR #322, opened 2021-05-04
Changed md5 to sha256 algorithm

nisarg1499:

### Changes
Detected MD5 hash algorithm which is considered insecure. MD5 is not collision resistant and is therefore not suitable as a cryptographic signature. Use SHA256 or SHA3 instead.

md5.Sum() is changed to sha256.Sum256()


sonasingh46:
Again, there is a trade-off and we need to take a call. SHA-256 is considerably slower than md5. If we strictly need the capabilities of SHA-256, we should go with this.
Let us dicsuss.

nisarg1499:
> Again, there is a trade-off and we need to take a call. SHA-256 is considerably slower than md5. If we strictly need the capabilities of SHA-256, we should go with this.
> Let us dicsuss.

@sonasingh46 Prometheus uses xxhash. We can use xxhash and it's faster than md5.

nisarg1499:
@sonasingh46 @mittachaitu @prateekpandey14 
Should we go with xxhash? Or md5 is fine for now?
Prometheus and InfluxDB uses xxhash and it's faster than md5. I am open to other suggestions.

sonasingh46:
@nisarg1499 -- xxhash makes sense to me. Let us go ahead with this. The  capabilities that we need from the hash function is : 
- It should be fast
- Low probability of collision. 
So xxhash vs md5 -- xxhash makes more sense in this scenario.

( Plus, I see that in places where this hash is used, there is no mechanism to deal with the hash collision, if at all. But that is out of the scope of this PR and can be thought of as part of the next PR. Just talking aloud )

nisarg1499:
> @nisarg1499 -- xxhash makes sense to me. Let us go ahead with this. The capabilities that we need from the hash function is :
> 
> * It should be fast
> * Low probability of collision.
>   So xxhash vs md5 -- xxhash makes more sense in this scenario.
> 
> ( Plus, I see that in places where this hash is used, there is no mechanism to deal with the hash collision, if at all. But that is out of the scope of this PR and can be thought of as part of the next PR. Just talking aloud )

@sonasingh46 Yaa I also saw that. There is no mechanism to deal with hash collision, and I would like to work on that also. I will create an issue and start a new PR. Is there any community channel link where I can ask questions regarding all these?

nisarg1499:
> I am good with [xxhash](https://github.com/cespare/xxhash), we can go ahead with it.
> 
> **Note**: Quickly verified the codebase to check effects if we change the hashing algorithm looks like there is no harm(since we are generating hash only once per volumereplica).

So I think we are good to go with xxhash. I'll add the xxhash algorithm.

nisarg1499:
@mittachaitu @sonasingh46 
Is this the latest function in use? Or the file where I have currently changed is in use? 
https://github.com/openebs/cstor-operators/blob/748dec6966f7ba8c3f449b4728a4155dfff3a71e/vendor/github.com/openebs/api/v2/pkg/util/hash.go#L28

Also, xxhash gives uint64 as output, so I was planning to convert that number into string using strconv.FormatUint. I hope this is fine.

mittachaitu:
> @mittachaitu @sonasingh46
> Is this the latest function in use? Or the file where I have currently changed is in use?
> https://github.com/openebs/cstor-operators/blob/748dec6966f7ba8c3f449b4728a4155dfff3a71e/vendor/github.com/openebs/api/v2/pkg/util/hash.go#L28
> 


@nisarg1499  we are using openebs/api Hash func at [volume_operations.go](https://github.com/openebs/cstor-operators/blob/748dec6966f7ba8c3f449b4728a4155dfff3a71e/pkg/controllers/cstorvolumeconfig/volume_operations.go#L840) and this can be updated to `Hash` function modified in this PR.

> Also, xxhash gives uint64 as output, so I was planning to convert that number into string using strconv.FormatUint. I hope this is fine.

Yes, you can convert.

nisarg1499:
@shubham14bajpai @mittachaitu 
xxhash changes done. Please review. I have kept base 10 as of now in strconv.FormatUint(). But in future if we want more security, we can change the base to other number.

nisarg1499:
> @nisarg1499 I forgot to ask one thing, I know that `md5sum` always produces (guarantee) 32 chars output, is it the same in `xxhash` case? Since we have [dependency](https://github.com/openebs/libcstor/blob/10c71becfc40d830f5d8c3449875212bbbbea246/include/zrepl_prot.h#L50) in other repos w.r.t the length of the checksum output

No xxhash does not produce 32 length string. It produces an uint64 of length 20. If we want we can append some random numbers to the output of xxhash.

sonasingh46:
@nisarg1499 -- Appending random numbers beats the purpose of hash value. 
@mittachaitu -- Why is the length of hash output a limitation?

nisarg1499:
> @nisarg1499 -- Appending random numbers beats the purpose of hash value.
> @mittachaitu -- Why is the length of hash output a limitation?

But we have the first 20 length unique output generated by hash. Then also appending the random numbers will beat the purpose?

sonasingh46:
>But we have the first 20 length unique output generated by hash. Then also appending the random numbers will beat the purpose?

That's where I asked the second question. If the 20 lengths unique output does the job why is the restriction for 32 chars. 
Plus, now I wonder how xxhash has a low collision rate compared to md5 (20 chars vs 32 chars)? Is it so. I do not know. Will need to check on this.

Appending random numbers will mean -- just to fit the code at the moment but serves no purpose. 
So either we get a 32 chars output ( completely hash ) or fix the limitation of 32 chars. 

@mittachaitu And as far as collision is concerned -- the code should tolerate the collisions irrespective of the algorithm being used.

nisarg1499:
> > But we have the first 20 length unique output generated by hash. Then also appending the random numbers will beat the purpose?
> 
> That's where I asked the second question. If the 20 lengths unique output does the job why is the restriction for 32 chars.
> Plus, now I wonder how xxhash has a low collision rate compared to md5 (20 chars vs 32 chars)? Is it so. I do not know. Will need to check on this.
> 
> Appending random numbers will mean -- just to fit the code at the moment but serves no purpose.
> So either we get a 32 chars output ( completely hash ) or fix the limitation of 32 chars.
> 
> @mittachaitu And as far as collision is concerned -- the code should tolerate the collisions irrespective of the algorithm being used.

@sonasingh46 I have not searched regarding the collision, but yeah I guess due to small length collision rate decreases.
Also @mittachaitu mentioned that we have some dependency [here](https://github.com/openebs/libcstor/blob/10c71becfc40d830f5d8c3449875212bbbbea246/include/zrepl_prot.h#L50) for 32 length string.

mittachaitu:
>  Is there any community channel link where I can ask questions regarding all these?

Hey @nisarg1499, We are available at [K8s slack](http://kubernetes.slack.com/) in the [#openebs-dev](https://kubernetes.slack.com/messages/openebs-dev) channel.  Discussions of this PR also happening [here](https://kubernetes.slack.com/archives/C015124PVTL/p1620848804218000) it started yesterday. Feel free to join [#openebs-dev](https://kubernetes.slack.com/messages/openebs-dev) channel

kmova:
@nisarg1499 @sonasingh46 @mittachaitu -- this was a good discussion thread. 

While `xxhash` sounds good and we should proceed towards that direction, could you please list down the test cases around upgrade that we need to verify once this change goes in?

mittachaitu:
> @nisarg1499 @sonasingh46 @mittachaitu -- this was a good discussion thread.
> 
> While `xxhash` sounds good and we should proceed towards that direction, could you please list down the test cases around upgrade that we need to verify once this change goes in?


Browsed through the cStor codebase to figure out breaking changes with different `hash` algorithms but **didn't find any such breaking changes**(Suspected there might be some issue when **replica registers with the target** (or) during **volume-replica scale down case** but we are handling them correctly so no issues). 

**Do we need to handle upgrade**:
No, not required.

**Following are the scenario tested in the K8s cluster**:
- Provisioned CSPC pool and busy-box application with single storage replica on top of provisioned cStor pool(at this time volume-replica controller is using md5sum hash algo).
- Updated the pool-manager-image to use xxhash algorightm.
- Scaled-up cStorVolume to hold 2 storage replicas, operation was successful and following is the CV content
```sh
 spec:
    ...
    ...
    ...
    replicaDetails:
      knownReplicas:
        "12784243745400926841": "4744040740243904044"    // Here key is generated using xxhash algo
        8C34945E4C524664DD9BF8553C25D154: "13207091750633283641" // Here key is generated using md5sum algo
  status:
    ...
    ...
    phase: Healthy
    replicaDetails:
      knownReplicas:
        "12784243745400926841": "4744040740243904044"
        8C34945E4C524664DD9BF8553C25D154: "13207091750633283641"
```
Some target logs:
```
2021-05-18/17:52:34.929 istgt_lu_add_unit :1792: u#10.140672308868864: trusty replica key {12784243745400926841} and zvol guid {4744040740243904044}

2021-05-18/17:52:34.929 istgt_lu_add_unit :1792: u#10.140672308868864: trusty replica key {8C34945E4C524664DD9BF8553C25D154} and zvol guid {13207091750633283641}
...
...
...
2021-05-18/17:52:36.219 update_replica_ent:1501: rm#-717228288.7     : replica(4744040740243904044) connected successfully from 10.244.4.9:3232 rep: 2

2021-05-18/17:52:37.234 update_replica_ent:1653: rm#-717228288.7     : replica(12784243745400926841:4744040740243904044) connected to target with needs_update: 0 can_be_trusty: 1
...
2021-05-18/17:52:40.225 update_replica_ent:1501: rm#-717228288.7     : replica(13207091750633283641) connected successfully from 10.244.5.13:3232 rep: 2

2021-05-18/17:52:41.247 update_replica_ent:1653: rm#-717228288.7     : replica(8C34945E4C524664DD9BF8553C25D154:13207091750633283641) connected to target with needs_update: 0 can_be_trusty: 1

2021-05-18/17:52:41.247 update_volstate   :1186: rm#-717228288.7     : volume(pvc-822cb05b-a146-41a7-bffb-786aeadccbe0) is ready for IOs now.. io_seq(10) healthy_replica(0) degraded_replica(2)
```
- Again, scaled-up storage replicas to 3(Operation was successful).
- Now, Scaled down the storage replicas to 2 and to 1(the operation was successful).

**Note**: We can use any hash algorithm if it produces output characters length <= 32 chars.

sonasingh46 (approved):
LGTM. Thank you
Alt Text

pkg/util/hash/hash.go
```diff
@@ -35,8 +35,8 @@ func Hash(obj interface{}) (string, error) {
 	if err != nil {
 		return "", errors.Wrapf(err, "failed to convert the object to json encoded format")
 	}
-	hashBytes := md5.Sum(jsonEncodedValues)
-	return hex.EncodeToString(hashBytes[:]), nil
+	hashBytes := xxhash.Sum64(jsonEncodedValues)
+	return strconv.FormatUint(hashBytes, 10), nil
```
mittachaitu, line 39: Magic number `10` over here is something like in which format output string is required. 
For example: 
- number 2 states give a string for a given hash in form of binary.
  example output: `100000110001010110110000101101011001101101111101101010101111100`
- number 8 states give a string for a given hash in form of octal
  example output: `1462530262414137617526`
- number 10 states give a string for a given hash in form of decimal
   example output: `14747893672817139542`
- number 16 states give a string for a given hash in form of hexadecimal
   example output: `ccab0b28617f1f56`

mittachaitu (commented):
@nisarg1499 can we update this [line](https://github.com/openebs/cstor-operators/blob/109066ef88c9ea0e3f83dec28da1bba04308264e/pkg/controllers/cstorvolumeconfig/volume_operations.go#L840) to use the hash from updated func(i.e pkg/util/hash) instead of using from openebs/api repo?

nisarg1499:
> @nisarg1499 can we update this [line](https://github.com/openebs/cstor-operators/blob/109066ef88c9ea0e3f83dec28da1bba04308264e/pkg/controllers/cstorvolumeconfig/volume_operations.go#L840) to use the hash from updated func(i.e pkg/util/hash) instead of using from openebs/api repo? So that it will be using the same hash func even in scaling operation.

@mittachaitu Are you suggesting to use xxhash.Sum64 instead of util.Hash of openebs/api in [this code line](https://github.com/openebs/cstor-operators/blob/109066ef88c9ea0e3f83dec28da1bba04308264e/pkg/controllers/cstorvolumeconfig/volume_operations.go#L840)

pkg/util/hash/hash.go
nisarg1499: Yes, the second parameter is the base in which we want the output. I kept it 10 so that we can get in form of decimal. Also the xxhash algorithm output is in decimal format.

mittachaitu:
> > @nisarg1499 can we update this [line](https://github.com/openebs/cstor-operators/blob/109066ef88c9ea0e3f83dec28da1bba04308264e/pkg/controllers/cstorvolumeconfig/volume_operations.go#L840) to use the hash from updated func(i.e pkg/util/hash) instead of using from openebs/api repo? So that it will be using the same hash func even in scaling operation.
> 
> @mittachaitu Are you suggesting to use xxhash.Sum64 instead of util.Hash of openebs/api in [this code line](https://github.com/openebs/cstor-operators/blob/109066ef88c9ea0e3f83dec28da1bba04308264e/pkg/controllers/cstorvolumeconfig/volume_operations.go#L840)

Yes

mittachaitu (approved):
Thanks for the PR @nisarg1499 ...
Changes are good

PR #323, opened 2021-05-04
JoinHostPort added

nisarg1499:

### Changes
Replaced fmt.Sprintf with net.JoinHostPort, because fmt.Sprintf will cause issues in IPv6. Best practice is to use net.JoinHostPort

### Results
Fixed  SemGrep complaint to use JoinHostPort instead of Sprintf for network address construction.


sonasingh46 (approved):
lgtm

mittachaitu (approved):
changes are good

PR #324, opened 2021-05-04
Weak random number generator replaced

nisarg1499:

### Changes
Currently in many files, weak random number generator(math/rand) is being used. Replaced it to crypto/rand.

### DIFF
--- a/pkg/debug/types.go
+++ b/pkg/debug/types.go
@@ -17,7 +17,7 @@ limitations under the License.
 package debug
 
 import (
-	"math/rand"
+	"crypto/rand"
 	"net/http"
 	"net/url"
 	"time"
--- a/pkg/controllers/cstorvolumeconfig/volume_operations.go
+++ b/pkg/controllers/cstorvolumeconfig/volume_operations.go
@@ -18,8 +18,8 @@ package cstorvolumeconfig
 
 import (
 	"context"
+	"crypto/rand"
 	"fmt"
-	"math/rand"
 	"strings"
 	"time"
 
--- a/pkg/controllers/testutil/zcmd/zpool/add.go
+++ b/pkg/controllers/testutil/zcmd/zpool/add.go
@@ -17,8 +17,8 @@ limitations under the License.
 package zpool
 
 import (
+	"crypto/rand"
 	"fmt"
-	"math/rand"
 	"strings"
 	"time"
 


arcolife:
Hey @nisarg1499 I ended up here while browsing the repo issues and PRs

You might wanna change more than just the import path to replace `math/rand` with `crypto/rand`. Take a look at https://golang.org/pkg/crypto/rand/ and you'd know that the functions differ vastly from https://golang.org/pkg/math/rand/

nisarg1499:
Hey @arcolife 
Actually I saw that, and I was studying about the functions. I don't know if we will get all current functionalities by `crypto/rand`, but still evaluating the difficulties in using this package as per our needs.
Thanks for figuring out.

sonasingh46:
@nisarg1499 -- Thank you for the PR.
Have you tested out by building the project and running tests locally after this change? 
Also if I am not wrong, the crypt/rand is more costly than math/rand. I am trying to see if we really need to use crypto/rand. 

Just a brief context on why this random stuff is used here : 
As part of a storage pool provisioning, multiple storage pools can get provisioned ( each on one node in a k8s cluster ) and then to create a volume -- the storage pools are picked up. ( For example, if you have 5 storage pools, and want to create volume with 3 replicas -- any 3 out of those 5 should get picked up )
Now the randomness is done so that any 3 pool is picked up for the following reasons: 
- The pools can be utilized evenly
- Do not dictate a bias(by the system) on selecting any first 3 pools or something like that. 

Open to discussion and learn more. ( Feel free to correct me ) 

cc @vishnuitta

nisarg1499:
> @nisarg1499 -- Thank you for the PR.
> Have you tested out by building the project and running tests locally after this change?
> Also if I am not wrong, the crypt/rand is more costly than math/rand. I am trying to see if we really need to use crypto/rand.
> 
> Just a brief context on why this random stuff is used here :
> As part of a storage pool provisioning, multiple storage pools can get provisioned ( each on one node in a k8s cluster ) and then to create a volume -- the storage pools are picked up. ( For example, if you have 5 storage pools, and want to create volume with 3 replicas -- any 3 out of those 5 should get picked up )
> Now the randomness is done so that any 3 pool is picked up for the following reasons:
> 
> * The pools can be utilized evenly
> * Do not dictate a bias(by the system) on selecting any first 3 pools or something like that.
> 
> Open to discussion and learn more. ( Feel free to correct me )
> 
> cc @vishnuitta

I am looking into this. Will revert back with some information.

pkg/debug/types.go
```diff
@@ -17,7 +17,7 @@ limitations under the License.
 package debug
 
 import (
-	"math/rand"
+	"crypto/rand"
```
prateekpandey14, line 20: The math/random returns a new pseudo-random source seeded with the given value, and crypto/rand does not depend on a seed value.

So corresponding usage below in function needs to be changed as well.

sonasingh46:
@nisarg1499 -- At this point, we should be using the `math/rand` for a while ( KISS principle )

Reasons : 
- Most cases the random shuffling is among 3 pool instances only. 
https://github.com/openebs/cstor-operators/blob/master/pkg/controllers/cstorvolumeconfig/volume_operations.go#L602
So it does not make much diff when the input size is this small.

- Again, not dealing with any security stuff. 

cc @kmova @mittachaitu @prateekpandey14 

If any of you feel, this should still make in, feel free to reopen the PR.  Closing it as of now.

PR #332, opened 2021-05-12
fix(cvc): set the bind address using a go flag in cvc server config

prateekpandey14:

**What this PR does ?**
- Remove the `entrypoint.sh` script in Dockerfile , used to set the bind address for cvc api server (backup/restore endpoints), reset/overwrite the other flags passed as args to the deployment container.

- Updated the cvc operator deployment spec args to include the `--bind` flag to set the server IP using downward API `status.podIP`


sonasingh46 (approved):
Alt Text

pkg/controllers/cstorvolumeconfig/start.go
```diff
@@ -33,10 +29,12 @@ import (
 	server "github.com/openebs/cstor-operators/pkg/server"
 	cvcserver "github.com/openebs/cstor-operators/pkg/server/cstorvolumeconfig"
 	"github.com/openebs/cstor-operators/pkg/snapshot"
+	"github.com/pkg/errors"
 	kubeinformers "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog"
 )
```
mittachaitu, line 39: Can we add a `klog.fatal` if [bind address](https://github.com/openebs/cstor-operators/blob/a08d9ae28665faf6e2e8416f7aa57750e54d5d76/pkg/controllers/cstorvolumeconfig/start.go#L174) is empty? This crash is required in a case where user doesn't specify `--bind` argument then the server start on pods [localhost](https://github.com/openebs/cstor-operators/blob/a08d9ae28665faf6e2e8416f7aa57750e54d5d76/pkg/server/config.go#L70) then other services in the cluster can not access the back/restore service.

This is required now since we are changing the way of passing arguments earlier it is always guaranteed that the entry script will pass value but now we are passing from YAML so for the safer side let's add a check and fatal if the value doesn't provide.

mittachaitu (commented):
provided minor comments

pkg/controllers/cstorvolumeconfig/start.go
prateekpandey14: Although since is going to be set via flag , its going to be set always using downward api , but added the fatal as well.

build/cvc-operator/Dockerfile.arm64
```diff
@@ -41,5 +41,5 @@ LABEL org.label-schema.build-date=$DBUILD_DATE
 LABEL org.label-schema.vcs-url=$DBUILD_REPO_URL
 LABEL org.label-schema.url=$DBUILD_SITE_URL
 
-ENTRYPOINT entrypoint.sh
```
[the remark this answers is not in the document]
prateekpandey14: have removed the arm dockerfile , no longer needed

mittachaitu (approved):
changes are good

PR #336, opened 2021-05-21
refact(charts): make cleanup pre hook image configurable

shubham14bajpai:

This PR makes the cleanup pre delete hook job image configurable for air-gap setups

### DIFF
--- a/deploy/helm/charts/README.md
+++ b/deploy/helm/charts/README.md
@@ -115,6 +115,8 @@ helm install openebs-cstor openebs-cstor/cstor --namespace openebs --create-name
 | admissionServer.resources | object | `{}` | Admission webhook pod resources |
 | admissionServer.securityContext | object | `{}` | Admission webhook security context |
 | admissionServer.tolerations | list | `[]` | Admission webhook tolerations |
+| cleanup.image.registry | string | `nil` |  cleanup pre hook image registry |
+| cleanup.image.repository | string | `"bitnami/kubectl"` |  cleanup pre hook image repository |
 | csiController.annotations | object | `{}` | CSI controller annotations |
 | csiController.attacher.image.pullPolicy | string | `"IfNotPresent"` | CSI attacher image pull policy |
 | csiController.attacher.image.registry | string | `"k8s.gcr.io/"` |  CSI attacher image registry |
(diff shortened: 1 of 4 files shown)


deploy/helm/charts/templates/cleanup-webhook.yaml
```diff
@@ -26,10 +26,10 @@ spec:
         - name: kubectl
           {{- /* bitnami maintains an image for all k8s versions */}}
           {{- /* see: https://hub.docker.com/r/bitnami/kubectl */}}
-          image: {{ printf "%s/%s:%s.%s" "bitnami" "kubectl" .Capabilities.KubeVersion.Major $kubeMinor }}
+          image: "{{ .Values.cleanup.image.registry }}{{ .Values.cleanup.image.repository }}:{{ .Capabilities.KubeVersion.Major }}.{{ $kubeMinor }}"
```
prateekpandey14, line 29: Same comment as openebs/charts PR , tag can be different as well.

deploy/helm/charts/values.yaml
```diff
@@ -238,3 +238,11 @@ analytics:
...
   # Specify in hours the duration after which a ping event needs to be sent.
   pingInterval: "24h"
+
+cleanup:
+  image:
+    # Make sure that registry name end with a '/'.
+    # For example : quay.io/ is a correct value here and quay.io is incorrect
+    registry:
+    repository: bitnami/kubectl
+    tag:
```
sonasingh46, line 248: Why the tag is empty? Are we going to use `latest` always?

deploy/helm/charts/README.md
```diff
@@ -115,6 +115,8 @@ helm install openebs-cstor openebs-cstor/cstor --namespace openebs --create-name
 | admissionServer.resources | object | `{}` | Admission webhook pod resources |
 | admissionServer.securityContext | object | `{}` | Admission webhook security context |
 | admissionServer.tolerations | list | `[]` | Admission webhook tolerations |
+| cleanup.image.registry | string | `nil` |  cleanup pre hook image registry |
```
sonasingh46, line 118: Question(n_p) : 
`nil` vs `""`
( I see that existing code uses `nil` )

deploy/helm/charts/values.yaml
shubham14bajpai: The tag is determined in the job template itself using the helm Capabilities to determine the k8s version, which will be same as the tag

deploy/helm/charts/README.md
shubham14bajpai: The existing values.yaml uses empty fields so I continued with that

prateekpandey14 (approved):

PR #346, opened 2021-06-18
fix(cleanup-job): add OR clause to prevent job from failing

prateekpandey14:

Cleanup job fails if the validatingwebhookconfiguration is deleted beforehand. This stalls helm chart remove. Eventually leads to timeout. Fixed it with a OR clause.


niladrih (approved):
LGTM

PR #354, opened 2021-07-02
fix(istgt config): add TCP User Timeout setting in istgt configuration

mittachaitu:
This commit adds a new setting called **TCPUserTimeout** in istgt configuration.
The default value of TCPUserTimeout is configured to 120 seconds which means
the maximum amount of time in seconds that transmitted data may remain unacknowledged,
or buffered data may remain untransmitted(due to zero window size) before
TCP will forcibly close the corresponding connection and returns an error event
to application.
    
**Why is this change needed**:
The current implementation of cStor-istgt allows only one initiator
to consume a LUN(volume) at a time. When an initiator machine is
abruptly powered off the connection is no more active between client & target.
To know about client connection status istgt configures TCP_USER_TIMEOUT
to 120 seconds during login time.
    
**NOTE**:
TCP_USER_TIMEOUT will helpful only if there is packet flow from istgt.
Here, if the connection remains idle for 20 seconds then istgt will send
Nop-In request to client(to check availability). If a client is available
things will work smoothly else if the packet is unacknowledged for 120 seconds
then the connection will be dropped.

**Note To Reviewers**:
- This can be merged after approving istgt [PR](https://github.com/openebs/istgt/pull/356).
- If an application node was powered off abruptly then it will take ~120 seconds for cstor-istgt
  to allow the new connection.
- If configuration is not provided then istgt will set default value to 120

### DIFF
--- a/pkg/controllers/volume-mgmt/volume/volume.go
+++ b/pkg/controllers/volume-mgmt/volume/volume.go
@@ -50,6 +50,7 @@ var (
   MediaDirectory "/mnt"
   Timeout 60
   NopInInterval 20
+  TCPUserTimeout 120
   MaxR2T 16
   DiscoveryAuthMethod None
   DiscoveryAuthGroup None


shubham14bajpai (approved):
LGTM

prateekpandey14 (approved):

PR #376, opened 2021-08-26
chore(operator): bump k8s csi sidecars images

prateekpandey14:
Update the k8s csi sidecar images to support minimum k8s 1.20.x version.

- CSI Provisioner:
k8s.gcr.io/sig-storage/csi-provisioner:v3.0.0

- CSI Resizer:
k8s.gcr.io/sig-storage/csi-resizer:v1.2.0

- CSI Node Driver Registrar
k8s.gcr.io/sig-storage/csi-node-driver-registrar:v2.3.0

**Note**: Keeping  the Snapshotter version as same as earlier , since the latest version only support the V1 snaphot apis , and  cloud k8s distribution, openshift etc does not supported the v1 snapshot APIs yet.

With current snapshotter 3.0.3 version  we have advantage that it will support both v1beta1/v1 apis, which will work in most of the cloud/openshift

### DIFF
--- a/deploy/csi-operator.yaml
+++ b/deploy/csi-operator.yaml
@@ -967,7 +967,7 @@ spec:
       serviceAccount: openebs-cstor-csi-controller-sa
       containers:
         - name: csi-resizer
-          image: k8s.gcr.io/sig-storage/csi-resizer:v1.1.0
+          image: k8s.gcr.io/sig-storage/csi-resizer:v1.2.0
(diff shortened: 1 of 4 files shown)


kmova (approved):

kmova:
@prateekpandey14 -- do we need to also update the helm chart accordingly?

kmova (approved):

PR #381, opened 2021-09-07
update(yaml): update priority class name for csi driver

shovanmaity:
use cstor in priority class name to avoid naming clash between different storage engines.

### DIFF
--- a/deploy/csi-operator.yaml
+++ b/deploy/csi-operator.yaml
@@ -923,7 +923,7 @@ roleRef:
 apiVersion: scheduling.k8s.io/v1
 kind: PriorityClass
 metadata:
-  name: openebs-csi-controller-critical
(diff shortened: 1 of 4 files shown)


mittachaitu (approved):
LGTM

deploy/csi-operator.yaml
```diff
@@ -923,15 +923,15 @@ roleRef:
 apiVersion: scheduling.k8s.io/v1
 kind: PriorityClass
 metadata:
-  name: openebs-csi-controller-critical
+  name: openebs-cstor-csi-controller-critical
```
kmova, line 926: Can we update the same here: https://github.com/openebs/cstor-operators/blob/master/deploy/csi-operator.yaml#L926

prateekpandey14 (approved):

kmova (approved):

PR #389, opened 2021-09-18
chore(ci): enable ndm change detection

kmova:
Signed-off-by: kmova <kiran.mova@mayadata.io>


kmova:
should be merged after #387

kmova:
Addressed by #390 

https://github.com/openebs/cstor-operators/blob/develop/deploy/cstor-operator.yaml#L3811

PR #349, opened 2021-06-23
fix(status): update cvr status on graceful corresponding pool-manager termination

prateekpandey14:
**What this PR does**

PR update CVR status to `offline` on graceful shutdown/termination of pool-manager pod/container deletion, eviction, scale down scenarios.

PR #395, opened 2021-10-09
fix(status) update cstor volume status to offline based on availability of target pods

### LINKED ISSUE #377: Update cstorVolume status based on the availablity of target pod
This CStorVolume status can be stale when the target pod is not in an available state. We can update the status of CStorVolume when a process in the container gets SIGINT/SIGKILL signal.

saltperfect:
Fixes https://github.com/openebs/cstor-operators/issues/377
I have tried to understand the project structure using documentation. 

Also I have referred to these two PR as suggested by @mittachaitu 
1. https://github.com/openebs/cstor-operators/pull/349
2. https://github.com/openebs/cstor-operators/pull/141 

I am yet to add testcases. This is an initial commit to get reviews and check if I have done the changes in correct file. 😸 Once I have nailed down which files, I can go ahead and improve on other aspects of the PR.

*First time contributor*

### DIFF
--- a/pkg/controllers/volume-mgmt/handler.go
+++ b/pkg/controllers/volume-mgmt/handler.go
@@ -767,8 +767,9 @@ func (c *CStorVolumeController) markCVStatusToOffline() {
 		klog.Errorf("failed to fetch CV list, error: %v", err)
 		return
 	}
-	for _, cv := range cvList.Items {
+	for i := range cvList.Items {
 		// Checking if Valid Cstor
+		cv := cvList.Items[i]
 		if IsValidCStorVolumeMgmt(&cv) {
 			cv.Status.Phase = apis.CVStatusOffline
 			cv.Status.ReplicaStatuses = nil
--- a/pkg/controllers/volume-mgmt/handler.go
+++ b/pkg/controllers/volume-mgmt/handler.go
@@ -31,6 +31,7 @@ import (
 	"github.com/openebs/api/v2/pkg/apis/types"
 	clientset "github.com/openebs/api/v2/pkg/client/clientset/versioned"
 	"github.com/openebs/api/v2/pkg/util"
+	"github.com/openebs/cstor-operators/pkg/controllers/common"
 	"github.com/openebs/cstor-operators/pkg/controllers/volume-mgmt/volume"
 	"github.com/openebs/cstor-operators/pkg/version"
 	corev1 "k8s.io/api/core/v1"
@@ -756,4 +757,36 @@ func (c *CStorVolumeController) reconcileVersion(cv *apis.CStorVolume) (*apis.CS
 		return cvObject, nil
 	}
 	return cv, nil
+}
+
+// markCVStatusToOffline will fetch all the CV resources present
+// in etcd and mark it them CV.Status to Offline
+func (c *CStorVolumeController) markCVStatusToOffline() {
+	
+	cvList, err := c.clientset.CstorV1().CStorVolumes("").List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		klog.Errorf("failed to fetch CV list, error: %v", err)
+		return
+	}
+	cvID := os.Getenv(string(common.OpenEBSIOCSPIID))
+	for _, cv := range cvList.Items {
+		cv := cv // pin it
+		if string(cv.GetUID()) == cvID {
+			// If target container restarts or pod is deleted
+			// before even creating a cvr, then we can skip those cvr
+			if cv.Status.Phase == "" || cv.Status.Phase == apis.CStorVolumePhase(apis.CVRStatusInit) {
+				continue
+			}
+			// No CstorVolume enum exist, hence wrapping cvr status
+			cv.Status.Phase = apis.CStorVolumePhase(apis.CVRStatusOffline)
+
+			// updating cv status to offline 
+			_, err = c.clientset.CstorV1().CStorVolumes(cv.Namespace).Update(context.TODO(), &cv, metav1.UpdateOptions{})
+			if err != nil {
+				klog.Errorf("failed to update CVR: %s status to %s", cv.Name, apis.CVRStatusOffline)
+				continue
+			}
+			klog.Infof("status marked %s for CV: %s", apis.CStorVolumePhase(apis.CVRStatusOffline), cv.Name)
+		}
+	}
 }
--- a/pkg/controllers/volume-mgmt/handler.go
+++ b/pkg/controllers/volume-mgmt/handler.go
@@ -768,7 +768,7 @@ func (c *CStorVolumeController) markCVStatusToOffline() {
 		return
 	}
 	for i := range cvList.Items {
-		// Checking if Valid Cstor
+		// Check if it is cStorVolume related to this target
 		cv := cvList.Items[i]
 		if IsValidCStorVolumeMgmt(&cv) {
 			if cv.Status.Phase == "" {
--- a/pkg/controllers/volume-mgmt/handler.go
+++ b/pkg/controllers/volume-mgmt/handler.go
@@ -31,7 +31,6 @@ import (
 	"github.com/openebs/api/v2/pkg/apis/types"
 	clientset "github.com/openebs/api/v2/pkg/client/clientset/versioned"
 	"github.com/openebs/api/v2/pkg/util"
-	"github.com/openebs/cstor-operators/pkg/controllers/common"
 	"github.com/openebs/cstor-operators/pkg/controllers/volume-mgmt/volume"
 	"github.com/openebs/cstor-operators/pkg/version"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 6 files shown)


pkg/controllers/volume-mgmt/handler.go
```diff
@@ -757,3 +758,35 @@ func (c *CStorVolumeController) reconcileVersion(cv *apis.CStorVolume) (*apis.CS
...
+				continue
+			}
+			// No CstorVolume enum exist, hence wrapping cvr status
+			cv.Status.Phase = apis.CStorVolumePhase(apis.CVRStatusOffline)
+
+			// updating cv status to offline 
+			_, err = c.clientset.CstorV1().CStorVolumes(cv.Namespace).Update(context.TODO(), &cv, metav1.UpdateOptions{})
+			if err != nil {
+				klog.Errorf("failed to update CVR: %s status to %s", cv.Name, apis.CVRStatusOffline)
+				continue
```
RealHarshThakur: Updating an object could result in an error due to conflict errors(most common), in which case...I think an error should be returned here which implies , function signature should be modified to return an error  :
```suggestion
		return err
```
And then when you're calling it from the controller, you can check for the error and if the error exists, you can requeue the object.

pkg/controllers/volume-mgmt/handler.go
```diff
@@ -757,3 +758,35 @@ func (c *CStorVolumeController) reconcileVersion(cv *apis.CStorVolume) (*apis.CS
...
+func (c *CStorVolumeController) markCVStatusToOffline() {
+	
+	cvList, err := c.clientset.CstorV1().CStorVolumes("").List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		klog.Errorf("failed to fetch CV list, error: %v", err)
+		return
+	}
+	cvID := os.Getenv(string(common.OpenEBSIOCSPIID))
+	for _, cv := range cvList.Items {
+		cv := cv // pin it
```
RealHarshThakur: I don't see why this is needed?

pkg/controllers/volume-mgmt/handler.go
```diff
@@ -757,3 +758,35 @@ func (c *CStorVolumeController) reconcileVersion(cv *apis.CStorVolume) (*apis.CS
 	}
 	return cv, nil
 }
+
+// markCVStatusToOffline will fetch all the CV resources present
+// in etcd and mark it them CV.Status to Offline
+func (c *CStorVolumeController) markCVStatusToOffline() {
+
```
RealHarshThakur: Nit: 
```suggestion
```

pkg/controllers/volume-mgmt/runner.go
```diff
@@ -49,6 +49,15 @@ func (c *CStorVolumeController) Run(threadiness int, stopCh <-chan struct{}) err
 
 	klog.Info("Started CStorVolume workers")
 	<-stopCh
+	
+	klog.Info("changing the CV state to offline before shutting down")
+	// Changing the state of corresponding CV to Offline before shutting down.
+	// Similar as when pod is running and if you stopped kubelet it will make
+	// pod status unknown.
+	// NOTE: CV status will be updated to OFFLINE even cStor target
+	// container alone get restarted but in this case pool will be still in
```
mittachaitu:
```suggestion
	// container alone get restarted but in this case, target(istgt) will be still in
```

pkg/controllers/volume-mgmt/handler.go
```diff
@@ -757,3 +758,35 @@ func (c *CStorVolumeController) reconcileVersion(cv *apis.CStorVolume) (*apis.CS
...
+// markCVStatusToOffline will fetch all the CV resources present
+// in etcd and mark it them CV.Status to Offline
+func (c *CStorVolumeController) markCVStatusToOffline() {
+	
+	cvList, err := c.clientset.CstorV1().CStorVolumes("").List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		klog.Errorf("failed to fetch CV list, error: %v", err)
+		return
+	}
```
mittachaitu: Here, it is not required to list CStorVolume resources on all namespaces... We can do it in following two ways:
- List resources in `openebs` namespace, the value of namespace we get via `OPENEBS_NAMESPACE` environment variable, and update the status.phase --> [Offline](https://github.com/openebs/api/blob/6f796956fb7307ab7ec16f93a7990895d1f6ffd7/pkg/apis/cstor/v1/cstorvolume.go#L211-L212) phase and [status.replicaStatuses](https://github.com/openebs/api/blob/6f796956fb7307ab7ec16f93a7990895d1f6ffd7/pkg/apis/cstor/v1/cstorvolume.go#L85) --> nil. This update has to be done only if CStorVolume that you are processing is belongs to the shutting down pod, we can add this check as done [here](https://github.com/openebs/cstor-operators/blob/e296a68467e2d47214fa6b0f8ce62d85e71a0f16/pkg/controllers/volume-mgmt/controller.go#L111)(It will ensure this is right CStorVolume).

We have to update only one CStorVolume Status, since there is only one-on-one mapping between CStorVolume resource and cStor-target pod as described [here](https://github.com/openebs/cstor-operators/blob/develop/docs/developer-guide/cstor-volume.md#resource-organisation)

pkg/controllers/volume-mgmt/handler.go
```diff
@@ -757,3 +758,35 @@ func (c *CStorVolumeController) reconcileVersion(cv *apis.CStorVolume) (*apis.CS
...
+	cvID := os.Getenv(string(common.OpenEBSIOCSPIID))
+	for _, cv := range cvList.Items {
+		cv := cv // pin it
+		if string(cv.GetUID()) == cvID {
+			// If target container restarts or pod is deleted
+			// before even creating a cvr, then we can skip those cvr
+			if cv.Status.Phase == "" || cv.Status.Phase == apis.CStorVolumePhase(apis.CVRStatusInit) {
+				continue
+			}
```
mittachaitu: CV status should not depends on CVR, CV has its own [states](https://github.com/openebs/api/blob/6f796956fb7307ab7ec16f93a7990895d1f6ffd7/pkg/apis/cstor/v1/cstorvolume.go#L204-L217) you can use from above constants.

pkg/controllers/volume-mgmt/handler.go
```diff
@@ -757,3 +758,35 @@ func (c *CStorVolumeController) reconcileVersion(cv *apis.CStorVolume) (*apis.CS
...
+			// No CstorVolume enum exist, hence wrapping cvr status
+			cv.Status.Phase = apis.CStorVolumePhase(apis.CVRStatusOffline)
+
+			// updating cv status to offline 
+			_, err = c.clientset.CstorV1().CStorVolumes(cv.Namespace).Update(context.TODO(), &cv, metav1.UpdateOptions{})
+			if err != nil {
+				klog.Errorf("failed to update CVR: %s status to %s", cv.Name, apis.CVRStatusOffline)
```
mittachaitu:
```suggestion
				klog.Errorf("failed to update CV: %s status to %s", cv.Name, apis.CVStatusOffline)
```

mittachaitu (commented):
Provided few comments PTAL

pkg/controllers/volume-mgmt/handler.go
saltperfect: Done.

pkg/controllers/volume-mgmt/handler.go
saltperfect: openebs/api version was outdated. Hence was not able to field in the package. Update to https://pkg.go.dev/github.com/openebs/api/v2/pkg/apis/cstor/v1  v2.4.0

pkg/controllers/volume-mgmt/handler.go
saltperfect: I have done the suggested changes. You can take a look now. Let me know if you see any more improvements. 
You suggested changing phase and replicaStatus. There are other fields such as [conditions](https://github.com/openebs/api/blob/6f796956fb7307ab7ec16f93a7990895d1f6ffd7/pkg/apis/cstor/v1/cstorvolume.go#L102) . Why do they don't need to get updated? Is that something the openebs control plane take care off?

mittachaitu: As of now, when a CV is undergoing resize operation conditions are populated and it will be removed only when [resize](https://github.com/openebs/cstor-operators/blob/develop/docs/tutorial/volumes/resize.md#how-to-expandresize-cstor-volume) operation is succeeded and conditions are mostly used for day-2-operations in cStor. Pod shutting down is not a day-2-operation so we are not updating them.

pkg/controllers/volume-mgmt/handler.go
```diff
@@ -757,3 +757,32 @@ func (c *CStorVolumeController) reconcileVersion(cv *apis.CStorVolume) (*apis.CS
...
+	cvList, err := c.clientset.CstorV1().CStorVolumes(openebsNameSpace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		klog.Errorf("failed to fetch CV list, error: %v", err)
+		return
+	}
+	for i := range cvList.Items {
+		// Checking if Valid Cstor
+		cv := cvList.Items[i]
+		if IsValidCStorVolumeMgmt(&cv) {
+			cv.Status.Phase = apis.CVStatusOffline
```
mittachaitu, line 777: Let's add check only if phase is non-empty(to maintain in sync with cvr & cspc status updation
```suggestion
                        if cv.Status.Phase == "" {
                              return
                        }
			cv.Status.Phase = apis.CVStatusOffline
```

mittachaitu (commented):
One more comment... Did you test the changes manually/automation way?

pkg/controllers/volume-mgmt/handler.go
saltperfect: I think you meant, If the phase is empty then continue. instead return right? [cvr commit](https://github.com/openebs/cstor-operators/pull/349/commits/d82e6ddfe20bda7f5005a03d2066fe928cd385c7#diff-214b27e3b1e501677482b660967b0f500018e110cae11d3f2505642d27fba096R723)

saltperfect: Also haven't test the changes yet. I have to setup and do that as well. Any documentation I can look at to test my changes?

mittachaitu: > I think you meant, If the phase is empty then continue. instead return right? 

No In CV case we can `return` since there is only a one-one mapping between cStor target pod and CStorVolume, if execution is happening here then we already found the right CV so not required to continue on others.

mittachaitu: > Also haven't test the changes yet. I have to setup and do that as well. Any documentation I can look at to test my changes?

Well!! 
- Please go through [prerequisites](https://github.com/openebs/cstor-operators/blob/develop/docs/developer-guide/devel.md#prerequisites) for the dev environment( Note: You can have the latest Go version also).
- Build the cStor-volume image using `make volume-manager-image` for building all images you can click [here](https://github.com/openebs/cstor-operators/blob/develop/docs/developer-guide/devel.md#build-the-images) 
- Once the image is built successfully, launch [minikube](https://minikube.sigs.k8s.io/docs/start/) cluster and follow [quick_start](https://github.com/openebs/cstor-operators/blob/develop/docs/quick.md#quickstart) doc to provision volume.
  Note: Minikube cluster might not have external disks, you can enable [sparse file count](https://github.com/openebs/charts/blob/ab0e45c96cd73f2f45c4d5431d3e7fa0ec903642/cstor-operator.yaml#L5317) to create sparse based blockdevices and use them for creating pools.

Once the volume is provisioned successfully, Scale down the cStor target pod, pod name starts with `pvc-XXX-YYYY....` and you should see corresponding resource i.e CV should mark as `Offline`

pkg/controllers/volume-mgmt/handler.go
```diff
@@ -757,3 +757,32 @@ func (c *CStorVolumeController) reconcileVersion(cv *apis.CStorVolume) (*apis.CS
...
+			cv.Status.ReplicaStatuses = nil
+			
+			// updating cv status to offline
+			_, err = c.clientset.CstorV1().CStorVolumes(cv.Namespace).Update(context.TODO(), &cv, metav1.UpdateOptions{})
+			if err != nil {
+				klog.Errorf("failed to update CV: %s status to %s", cv.Name, apis.CVStatusOffline)
+				continue
```
mittachaitu: Here also, we found the CV that required to update and processing other CV's will not helpful so we can return
```suggestion
				return
```

pkg/controllers/volume-mgmt/handler.go
mittachaitu: You can also report if there is any issue related to dev setup in [#openebs-dev](https://kubernetes.slack.com/archives/C015124PVTL) channel.

pkg/controllers/volume-mgmt/handler.go
```diff
@@ -757,3 +757,35 @@ func (c *CStorVolumeController) reconcileVersion(cv *apis.CStorVolume) (*apis.CS
...
+// in etcd and mark it them CV.Status to Offline
+func (c *CStorVolumeController) markCVStatusToOffline() {
+	openebsNameSpace := os.Getenv("OPENEBS_NAMESPACE")
+	cvList, err := c.clientset.CstorV1().CStorVolumes(openebsNameSpace).List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		klog.Errorf("failed to fetch CV list, error: %v", err)
+		return
+	}
+	for i := range cvList.Items {
+		// Checking if Valid Cstor
```
mittachaitu:
```suggestion
		// Check if it is cStorVolume related to this target
```

mittachaitu (commented):
Looks good... Let's rollout PR from draft stage to PR

mittachaitu (approved):
Changes are good

prateekpandey14 (approved):
lgtm

PR #397, opened 2021-10-12
fix(helm): Make log verbosity configurable

ianroberts:
A number of components in the helm charts have hard-coded `--v=5` options which can produce overly-verbose logging. I recently submitted openebs/jiva-operator#161 to fix this in the Jiva chart and was asked by @prateekpandey14 to port the same changes here for cStor.

**What this PR does?**:

This PR makes various `--v` flags configurable via the values file/`--set`.  To maintain backwards compatibility the defaults in `values.yaml` remain at the previously hard-coded values but they can now be overridden via

- `csiController.logLevel` to change all the CSI controller containers to the same level in one go.  If this value is not set then we fall back to per-container settings, all of which default to "5"
  - `csiController.attacher.logLevel`
  - `csiController.provisioner.logLevel`
  - `csiController.resizer.logLevel`
  - `csiController.snapshotController.logLevel`
- `csiNode.logLevel` as above for the CSI node pod, though there is only one container-specific child value in this case:
  - `csiNode.driverRegistrar.logLevel` for the node pod driver registrar
- `cvcOperator.logLevel` for the CVC operator (default level 2)

---

**Edit**: following discussion this has now been changed so that the top-level per-pod settings provide the defaults in `values.yaml`, and the per-container settings can be supplied to override the default.

---

**Note** I have only added options for those cases where there is already a `--v` option specified in the helm template.  There may be other components that _could_ accept `--v` but where the chart does not currently pass one, but I don't use cStor myself (this PR is a port of changes I made to Jiva) and don't know the code so I've limited myself to the obviously non-breaking changes.

**Does this PR require any upgrade changes?**:

No

### DIFF
--- a/deploy/helm/charts/README.md
+++ b/deploy/helm/charts/README.md
@@ -122,19 +122,23 @@ helm install openebs-cstor openebs-cstor/cstor --namespace openebs --create-name
 | csiController.attacher.image.registry | string | `"k8s.gcr.io/"` |  CSI attacher image registry |
 | csiController.attacher.image.repository | string | `"sig-storage/csi-attacher"` |  CSI attacher image repo |
 | csiController.attacher.image.tag | string | `"v3.1.0"` | CSI attacher image tag |
+| csiController.attacher.logLevel | string | _unspecified_ | Override log level for CSI attacher container (1 = least verbose, 5 = most verbose) |
 | csiController.attacher.name | string | `"csi-attacher"` |  CSI attacher container name|
 | csiController.componentName | string | `"openebs-cstor-csi-controller"` | CSI controller component name |
+| csiController.logLevel | string | `"5"` |  Default log level for all CSI controller containers (1 = least verbose, 5 = most verbose) unless overridden for a specific container |
 | csiController.nodeSelector | object | `{}` |  CSI controller pod node selector |
 | csiController.podAnnotations | object | `{}` | CSI controller pod annotations |
 | csiController.provisioner.image.pullPolicy | string | `"IfNotPresent"` | CSI provisioner image pull policy |
 | csiController.provisioner.image.registry | string | `"k8s.gcr.io/"` | CSI provisioner image pull registry |
 | csiController.provisioner.image.repository | string | `"sig-storage/csi-provisioner"` | CSI provisioner image pull repository |
 | csiController.provisioner.image.tag | string | `"v3.0.0"` | CSI provisioner image tag |
+| csiController.provisioner.logLevel | string | _unspecified_ | Override log level for CSI provisioner container (1 = least verbose, 5 = most verbose) |
 | csiController.provisioner.name | string | `"csi-provisioner"` | CSI provisioner container name |
 | csiController.resizer.image.pullPolicy | string | `"IfNotPresent"` | CSI resizer image pull policy  |
 | csiController.resizer.image.registry | string | `"k8s.gcr.io/"` | CSI resizer image registry |
 | csiController.resizer.image.repository | string | `"sig-storage/csi-resizer"` |  CSI resizer image repository|
 | csiController.resizer.image.tag | string | `"v1.2.0"` | CSI resizer image tag |
+| csiController.resizer.logLevel | string | _unspecified_ | Override log level for CSI resizer container (1 = least verbose, 5 = most verbose) |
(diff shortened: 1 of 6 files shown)


ianroberts:
Sorry for the noise, I changed `version` in the wrong place...

deploy/helm/charts/README.md
```diff
@@ -122,26 +122,31 @@ helm install openebs-cstor openebs-cstor/cstor --namespace openebs --create-name
 | csiController.attacher.image.registry | string | `"k8s.gcr.io/"` |  CSI attacher image registry |
 | csiController.attacher.image.repository | string | `"sig-storage/csi-attacher"` |  CSI attacher image repo |
 | csiController.attacher.image.tag | string | `"v3.1.0"` | CSI attacher image tag |
+| csiController.attacher.logLevel | string | `"5"` |  CSI attacher container log level (1 = least verbose, 5 = most verbose)|
 | csiController.attacher.name | string | `"csi-attacher"` |  CSI attacher container name|
 | csiController.componentName | string | `"openebs-cstor-csi-controller"` | CSI controller component name |
+| csiController.logLevel | string | _unspecified_ |  Log level for all CSI controller containers (1 = least verbose, 5 = most verbose) - overrides the per-container defaults |
```
mittachaitu: Should It be the other way, like `csiController.logLevel` will be default values for all CSI controller but we can override this value by specifying per container level?
```suggestion
| csiController.logLevel | string | _unspecified_ |  Log level for all CSI controller containers (1 = least verbose, 5 = most verbose) - will get override by per-container defaults |
```

deploy/helm/charts/templates/csi-controller.yaml
```diff
@@ -29,7 +29,7 @@ spec:
           resources:
 {{ toYaml .Values.csiController.resources | indent 12 }}
           args:
-            - "--v=5"
+            - "--v={{ .Values.csiController.logLevel | default .Values.csiController.resizer.logLevel }}"
```
mittachaitu: IMO, it should be in below way... 
- If the configuration is not specified then high-level setting will be applied but if it is specified at container level then those settings should be applied
```suggestion
            - "--v={{  .Values.csiController.resizer.logLevel | default .Values.csiController.logLevel }}"
```

Simillarly, In other places and also open for discussion

mittachaitu (commented):
@ianroberts , Thanks for raising the PR, have provided a few comments

deploy/helm/charts/README.md
ianroberts: So the behaviour at the moment (matching my jiva-operator PR that has already been merged) is that if someone sets both `csiController.logLevel` _and_ `csiController.provisioner.logLevel`, then the former setting will override the latter.  If you want to change all the log levels except the provisioner then you'd have to set the `attacher`, `resizer` and `snapshotController` explicitly.  I agree it probably does make more sense the other way round but if I change it here I should change it in the jiva-operator as well to keep things consistent and it's already been released the current way in jiva-operator v3.0.3.

@prateekpandey14 what do you think, would it be OK to reverse the overriding order in jiva too?

deploy/helm/charts/templates/csi-controller.yaml
ianroberts: Agreed, but see [my comment on your other suggestion](https://github.com/openebs/cstor-operators/pull/397#discussion_r736404390)

deploy/helm/charts/README.md
mittachaitu: My take is to reverse the existing way since it is not appropriate... Since these are configuration parameters users should check README.md/Release notes before upgrading to the latest version. Let's wait for @prateekpandey14 and others suggestion also

prateekpandey14: I think @mittachaitu suggestion does make sense, we should be doing the other way. Similary we can update the jiva-operators charts as well, since we did take care of consistency with older chart.

ianroberts: Thanks, I've switched it round, rebased and force-pushed a new version.

mittachaitu (approved):
LGTM

prateekpandey14 (approved):
lgtm

mittachaitu (approved):
LGTM

Ab-hishek (approved):

PR #400, opened 2021-10-19
fix(pool): set canmount filesystem property to off for pool dataset

mittachaitu:
This PR does the following changes to the codebase:
- Configures `canmount` filesystem property to off for pool dataset. It also takes care of the existing pool to update their property once pools are upgraded to the latest version.
- Add integration test for verifying pool properties
- Add make target(`make integration-test`) for running integration tests.
- Updates the sanity test to run integration-test.
- Updates the integration `_test` files to use K8s client according to go.mod.


pkg/pool/operations/property.go
```diff
@@ -0,0 +1,126 @@
...
+		dsProperties.fsProperties[key] = value
+	}
+	ds.datasetsToProperties[datasetName] = dsProperties
+
+	return nil
+}
+
+// SetPoolProperties will configure required pool properties
+// NOTE: Advantage of using SetPoolProperties is it avoids calls to cstor-pool
+//		 executing zpool/zfs commands(most of them) will read from disk
```
prateekpandey14:
```suggestion
// NOTE: Advantage of using SetPoolProperties is to avoids calls to cstor-pool
// since executing zpool/zfs commands(most of them) will read from disk
```

prateekpandey14: nit : I think above code comment should be in `SetPoolFSPropertiesIfNot` not here ?

mittachaitu: Right, I will move it

pkg/pool/operations/property.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+		// Read the property from cstor-pool container
+		existingPropValue, err := oc.GetVolumePropertyValue(datasetName, property)
+		if err != nil {
+			return errors.Wrapf(err, "failed to get value of filesystem property %s of pool %s", property, datasetName)
+		}
+		if existingPropValue == desiredValue {
+			dsProperties.fsProperties[property] = desiredValue
+			continue
+		}
```
prateekpandey14, line 77: do we still need this , we already verifiled it above ?

prateekpandey14: ahh , ok we are trying to skip the same value, below ?

prateekpandey14 (commented):
changes and integration tests looks good, few comments though.

pkg/pool/operations/property.go
mittachaitu: No, Here we are updating in-memory properties... This is a case after restarting container in-memory will be empty, to fill in-memory we are updating this

pkg/pool/operations/property.go
mittachaitu: Updated code comments

ci/sanity/cspc.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: cstor.openebs.io/v1
```
shubham14bajpai: This file is not needed

pkg/controllers/cspi-controller/handler.go
```diff
@@ -448,15 +448,14 @@ func (c *CStorPoolInstanceController) addPoolProtectionFinalizer(
...
-	err := oc.SetCompression(poolName, compressionType)
+
+	// reconcile pool(fs & pool both) properties
+	err := oc.SetPoolProperties(cspi)
 	if err != nil {
 		c.recorder.Event(cspi,
 			corev1.EventTypeWarning,
-			"Pool "+string("FailedToSetCompression"),
-			fmt.Sprintf("Failed to set compression %s to the pool %s : %s", compressionType, poolName, err.Error()))
+			"Pool "+string("FailedToPoolProperties"),
```
shubham14bajpai:
```suggestion
			"Pool "+string("FailedToSetProperties"),
```

shubham14bajpai (approved):
LGTM, just some minor comments

ci/sanity/cspc.yaml
mittachaitu: removed this autogenerated file

pkg/controllers/cspi-controller/handler.go
mittachaitu: Update the string

prateekpandey14 (approved):

PR #423, opened 2022-04-15
[stable/cStor]: update cStor charts to 3.2.0 release

Ab-hishek:
This PR does the following changes:

Updates cStor charts to use new 3.2.0 release.


Ab-hishek:
@mittachaitu Do we need to update the kubernetes side car images also for the csi drivers?

mittachaitu (approved):
LGTM

mittachaitu:
> @mittachaitu Do we need to update the kubernetes side car images also for the csi drivers?

Usually we will bump them in the following scenarios:
- When the latest version of Kubernetes depreciates older images of CSI sidecar.
- When we bump the minimum required version for Kubernetes.

PR #426, opened 2022-06-17
Fix can't patch cvc

nobiit:


pkg/controllers/cstorvolumeconfig/volume_operations.go
```diff
@@ -935,10 +935,13 @@ func (c *CVCController) scaleDownVolumeReplicas(cvc *apis.CStorVolumeConfig) (*a
...
-				Delete(context.TODO(), cvrObj.Name, metav1.DeleteOptions{})
-			if err != nil {
-				return cvc, errors.Wrapf(err, "failed to delete cstorvolumereplica %s", cvrObj.Name)
+			if len(cvrObj.Name) > 0 { 
+				// TODO: Skip deletion if user deleted it manually (avoid getting stuck)
+				err = c.clientset.CstorV1().CStorVolumeReplicas(openebsNamespace).
+					Delete(context.TODO(), cvrObj.Name, metav1.DeleteOptions{})
+				if err != nil {
+					return cvc, errors.Wrapf(err, "failed to delete cstorvolumereplica %s", cvrObj.Name)
+				}
```
mittachaitu, line 944: Thanks for the PR @nobiit, Can we handle this case differently? Like after triggering deletion of CVR, can you check for error and if error is other than not found we can return error else case is good.

nobiit:
Great idea @mittachaitu . I am learning the architecture of cstor, give me a little more time.
P/s: I still can't fix the original error, so please hold this PR temporarily

nobiit:
Well, I'll close this issue since I didn't come across it again after much testing with Cstor. Overall, it's great :D

PR #436, opened 2022-09-27
fix(cvc-controller): add support for policy/v1 PodDisruptionBudget

niladrih:
Signed-off-by: Niladri Halder <niladri.halder26@gmail.com>

Changes:
- Add support for policy/v1 PodDisruptionBudget in the CVC controller. This is required because policy/v1beta1 is no longer supported in k8s v1.25.
- Upgrade go version to 1.19, update k8s libs to 0.25.2.
- Remove vendor directory.


build/cstor-webhook/cstor-webhook.Dockerfile
```diff
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-FROM golang:1.14.7 as build
+FROM golang:1.19.1 as build
```
vharsh, line 14: In Go 1.19, GO111MODULE env variable doesn't need to be set, please fix line#22

pkg/controllers/cstorvolumeconfig/controller_base.go
```diff
@@ -324,3 +331,28 @@ func (c *CVCController) processNextWorkItem() bool {
...
 	return true
 }
+
+func isV1PodDisruptionBudgetSupported(client discovery.DiscoveryInterface) bool {
+	const (
+		podDisruptionBudgetKind      = "PodDisruptionBudget"
+		podDisruptionBudgetStableApi = "policy/v1"
+	)
+
+	//Checking if the policy/v1 API contains the CronJob Kind
```
vharsh: nitpick:

```suggestion
	// Checking if the policy/v1 API contains the PodDisruptionBudget Kind
```

pkg/controllers/cstorvolumeconfig/controller_base.go
```diff
@@ -324,3 +331,28 @@ func (c *CVCController) processNextWorkItem() bool {
...
+	apiList, err := client.ServerResourcesForGroupVersion(podDisruptionBudgetStableApi)
+	if err != nil {
+		if k8serror.IsNotFound(err) {
```
vharsh: is the indentation okay? please re-run go-fmt on save for this file

pkg/controllers/cstorvolumeconfig/controller_base.go
```diff
@@ -324,3 +331,28 @@ func (c *CVCController) processNextWorkItem() bool {
...
 }
+
+func isV1PodDisruptionBudgetSupported(client discovery.DiscoveryInterface) bool {
+	const (
+		podDisruptionBudgetKind      = "PodDisruptionBudget"
+		podDisruptionBudgetStableApi = "policy/v1"
+	)
+
+	//Checking if the policy/v1 API contains the CronJob Kind
+	apiList, err := client.ServerResourcesForGroupVersion(podDisruptionBudgetStableApi)
```
vharsh: I guess this uses the discovery API, can we not pass a `metav1.ListOptions{}` to only get a specific kind of resource, this way we wouldn't have to iterate on it, from what I can recall it doesn't but please check if you've got time.

pkg/controllers/cstorvolumeconfig/pdb.go
```diff
@@ -76,6 +77,28 @@ func (c *CVCController) createPDB(poolNames []string, cspcName string) (*policy.
 		Create(context.TODO(), pdbObj, metav1.CreateOptions{})
 }
 
+// createV1PDB creates PDB for cStorVolumes based on arguments
+func (c *CVCController) createV1PDB(poolNames []string, cspcName string) (*policyv1.PodDisruptionBudget, error) {
+	// Calculate minAvailable value from cStorVolume replica count
+	//minAvailable := (cvObj.Spec.ReplicationFactor >> 1) + 1
```
vharsh: Why is this line commented out & doesn't look like a comment but a line of unused variable declaration?

pkg/controllers/cstorvolumeconfig/volume_operations.go
```diff
@@ -697,38 +733,49 @@ func (c *CVCController) getUpdatePDBForVolume(cvcObj *apis.CStorVolumeConfig) (s
 	if !isHAVolume(cvcObj) {
 		return "", nil
 	}
-	pdbObj, err := c.getOrCreatePodDisruptionBudget(getCSPC(cvcObj), cvcObj.Status.PoolInfo)
-	if err != nil {
-		return "", err
+
+	if useV1beta1PodDisruptionBudget {
+		pdbObj, err := c.getOrCreateV1beta1PodDisruptionBudget(getCSPC(cvcObj), cvcObj.Status.PoolInfo)
```
vharsh: nit: I personally don't like long named methods, if other PR reviewers feel the same way, maybe the names can be changed to `getOrCreatev1beta1PDB`.

pkg/controllers/replica-controller/runner_test.go
```diff
@@ -249,14 +250,14 @@ func (r TestRunner) RunCommandWithLog(common string, args ...string) ([]byte, er
 
 // TestGetterProcess mocks zpool get.
 func TestGetterProcess(*testing.T) {
-	defer os.Exit(0)
+	defer syscall.Exit(0)
```
vharsh, line 253: has this been changed as per IDE suggestion? I believe some golang version had a different behaviour towards `os.Exit` being called from tests

vharsh (commented):
minor change suggestions

pkg/controllers/cstorvolumeconfig/controller_base.go
niladrih: Oh, yeah, lol

pkg/controllers/cstorvolumeconfig/controller_base.go
niladrih: It's 8 spaces. It's correct. Github uses 8 spaces for tabs.

pkg/controllers/cstorvolumeconfig/pdb.go
niladrih: It was there to begin with on the v1beta1 function. So I've carried it forward, as any change there should also  be reflected on the v1 function.

pkg/controllers/cstorvolumeconfig/controller_base.go
niladrih: No, I didn't find an interface to query if a specific object is present.

build/cstor-webhook/cstor-webhook.Dockerfile
niladrih: Removed the env from all of the multi-arch dockerfiles

pkg/controllers/cstorvolumeconfig/controller_base.go
niladrih: Fixed this.

pkg/controllers/replica-controller/runner_test.go
niladrih: Yes.... that's why I'm using syscall. os.Exit() used in testing is an automatic failure.

pkg/controllers/cstorvolumeconfig/controller.go
```diff
@@ -330,13 +331,23 @@ func (c *CVCController) createVolumeOperation(cvc *apis.CStorVolumeConfig) (*api
...
+			}
+			addV1beta1PDBLabelOnCVC(cvc, pdbObj)
+		} else {
+			var pdbObj *policyv1.PodDisruptionBudget
+			pdbObj, err = c.getOrCreateV1PodDisruptionBudget(getCSPC(cvc), poolNames)
+			if err != nil {
+				return nil, errors.Wrapf(err,
+					"failed to create PDB for volume: %s", cvc.Name)
+			}
+			addV1PDBLabelOnCVC(cvc, pdbObj)
```
Ab-hishek: Do we need all these if...else statements? Should we simply change this to v1 api and update the docs specifying the minimum K8s version required by cStor to work? This way we won't be needing to support both the versions simultaneously.

niladrih: Fair enough ✔️

pkg/controllers/cstorvolumeconfig/volume_operations.go
niladrih: Reverted to the previous function name. Reverted changes which supported both policy/v1beta1 and policy/v1. Only support for policy/v1 will be there.

.github/workflows/build.yml
```diff
@@ -72,15 +72,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        kubernetes: [v1.20.6]
+        kubernetes: [v1.19.16, v1.25.0]
```
Ab-hishek: Why two versions?

Ab-hishek (approved):
LGTM

.github/workflows/build.yml
niladrih: To check if the build passed for kubernetes v1.19 and v1.25. It passed maybe because the volumes were not HA, and PDB object was not created. v1.19 should be removed. Good catch!

PR #437, opened 2022-09-28
docs(cspc): update CSPC Mirror docs

WillyRL:
In the Mirror RAID Docs on section #Disk Replacement, There's a line that I believe come from Stripe RAID Config docs


niladrih (approved):
Thank you for the contribution @WillyRL

Issue #434, opened 2022-08-11
Wrong level of identation when supplying an imagePullSecrets


jam-re:
When specifying a custom imagePullSecrets, the generated yaml is invalid due to the ident level specified in the template (should be 8 and is instead 2) :

Example :
https://github.com/openebs/cstor-operators/blob/v3.3.x/deploy/helm/charts/templates/cspc-operator.yaml#L77

```
cstor:
  enabled: true
  csiController.attacher.image.registry: my.private.registry/
  csiController.driverRegistrar.image.registry: my.private.registry/
  csiController.provisioner.image.registry: my.private.registry/
  csiController.resizer.image.registry: my.private.registry/
  csiController.resizer.image.registry: my.private.registry/
  csiController.snapshotter.image.registry: my.private.registry/
  csiNode.driverRegistrar.image.registry: my.private.registry/
  cspcOperator.cstorPool.image.registry: my.private.registry/
  cspcOperator.cstorPoolExporter.image.registry: my.private.registry/
  cspcOperator.image.registry: my.private.registry/
  cspcOperator.poolManager.image.registry: my.private.registry/
  cstorCSIPlugin.image.registry: my.private.registry/
  cvcOperator.image.registry: my.private.registry/
  cvcOperator.target.image.registry: my.private.registry/
  cvcOperator.volumeExporter.image.registry: my.private.registry/
  cvcOperator.volumeMgmt.image.registry: my.private.registry/
  admissionServer.image.registry: my.private.registry/
  imagePullSecrets:
    - name: my-private-registry
```

PR #453, opened 2023-11-01
fix incorrect docker secret handling #434

sjerman:
Fix a couple of issues in the Helm chart. See #434


prateekpandey14 (changes requested):
@sjerman you have to bump the chart version in `Charts.yaml` file.

sjerman:
Increased helm chart version from 3.5.0 -> 3.5.1

sjerman:
Looks like I need to signoff a couple of commits... Will do now.

